PR #13, opened 2021-05-31
Refactor & Fixes

rahmatrhd:


domain/constants.go
```diff
@@ -0,0 +1,5 @@
+package domain
+
+const (
+	AuthenticatedEmailHeaderKey = "X-Goog-Authenticated-User-Email"
```
StewartJingga, line 4: if only appeal package uses it should we consider keeping it inside the package instead of domain?
it could be a normal private constant in appeal package. wdyt?

provider/bigquery/bigquery.go
```diff
@@ -10,15 +10,13 @@ import (
 	"google.golang.org/api/option"
 )
 
-// Client is wrapper for bigquery client
-type Client struct {
+type bigQueryClient struct {
```
StewartJingga, line 13: if possible, since the package name is already bigquery, we could probably just use `client` instead of `bigQueryClient`.

but since it is a private struct, it should be okay since only the package knows about it, you can decide on this.

ref: https://blog.golang.org/package-names#TOC_3.

domain/approval.go
```diff
@@ -26,6 +26,10 @@ type Approval struct {
 	UpdatedAt time.Time `json:"updated_at"`
 }
 
+func (a *Approval) IsHumanApproval() bool {
```
StewartJingga, line 29: Pardon my lack of context, but would `IsManualApproval` better here?

provider/bigquery/bigquery.go
rahmatrhd: the reason I renaming it to `bigqueryClient` is that I created another client named `iamClient`, to make them look as they are in the same position as clients

domain/approval.go
rahmatrhd: that works, changing it 👍

domain/constants.go
rahmatrhd: yes currently it is only used in the appeal handler. This is actually more like a contract for application-wide, so I put it in the domain package instead. Also, there's a possibility that this constant will be used in other packages as well.
It is more proper to make this configurable actually, what do you think?

StewartJingga: usually my rule of thumb is if it is being shared across packages then domain would make sense.
If the possibility is pretty high then yeah it would probably okay to put it there.

But also another concern is that this constant is related to HTTP so imho keeping it as close to the handler using it would be better. wdyt?

rahmatrhd: Alright, we'll see how to position this properly as this project grows. For now let's move it closer to the related use case, moving it to appeal handler for now.

StewartJingga (approved):

PR #14, opened 2021-06-16
Metabase Provider

rahmatrhd:


provider/metabase/resource.go
```diff
@@ -0,0 +1,6 @@
+package metabase
+
+const (
+	ResourceTypeDatabase   = "database"
```
StewartJingga, line 4: I was wondering if these could be private variables? If they could, should we consider changing them? It could help understanding the code better, wdyt?

rahmatrhd: making it public because it is used in testing https://github.com/odpf/guardian/pull/15/files#diff-31406187a198c44feeb060c9668d0761c843da91498e6f8bd331663ef88ead70R24

StewartJingga: ah I see 🤔, okay then

StewartJingga (approved):

PR #28, opened 2021-08-03
GRPC

rahmatrhd:
- [x] Init grpc server
- [x] Get /providers
- [x] Post /providers
- [x] Put /providers/{id}
- [ ] Get /policies
- [ ] Post /policies
- [ ] Put /policies/{id}
- [ ] Get /resources
- [ ] Put /resources/{id}
- [ ] Post /appeals
- [ ] Get /appeals
- [ ] Get /appeals/approvals
- [ ] Post /appeals/{id}/approvals/{name}
- [ ] Put /appeals/{id}/cancel
- [ ] Put /appeals/{id}/revoke
- [ ] Get /appeals/{id}


api/handler/v1/guardian_service.go
```diff
@@ -0,0 +1,445 @@
...
+	"github.com/odpf/guardian/appeal"
+	"github.com/odpf/guardian/domain"
+	"github.com/odpf/guardian/policy"
+	"github.com/odpf/guardian/resource"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+type ProtoAdapter interface {
```
StewartJingga, line 18: would it be simpler if we use struct instead of interface for the adapter?

and to make it even more simpler, we could also remove the adapter and change these methods as a standalone functions, just like helper functions. 

since the adapter does not have any state and we only use it in v1 package, I think it should be doable. wdyt?
e.g.
```
package v1

func fromProviderProto(*pb.Provider) (*domain.Provider, error) {
  return
}
```

api/handler/v1/guardian_service.go
```diff
@@ -0,0 +1,445 @@
...
+	FromResourceProto(*pb.Resource) *domain.Resource
+	ToResourceProto(*domain.Resource) (*pb.Resource, error)
+
+	FromAppealProto(*pb.Appeal) (*domain.Appeal, error)
+	ToAppealProto(*domain.Appeal) (*pb.Appeal, error)
+	FromCreateAppealProto(*pb.CreateAppealRequest) ([]*domain.Appeal, error)
+	ToApprovalProto(*domain.Approval) (*pb.Approval, error)
+}
+
+type GuardianServiceServer struct {
```
StewartJingga, line 35: I realise our service protos have `Guardian` name in it so we can leave them as it.

But for the one that is written in this repo, like this `GuardianServiceServer`, does it make sense if we change it to GRPCServer instead?

StewartJingga: Or maybe if it also handles non-grpc, we can use `ServiceServer` to be more generic?

StewartJingga (commented):
added a few comments

api/handler/v1/guardian_service.go
rahmatrhd: Using interface so that we can create tests for both adapter and grpc service independently. This interface will help to create tests easier by using mock

StewartJingga (approved):

PR #34, opened 2021-08-12
feat: guardian cli

rahmatrhd:


cmd/root.go
```diff
@@ -4,16 +4,31 @@ import (
...
 func Execute() {
+	var rootCmd = &cobra.Command{
+		Use: "guardian",
+	}
+
+	protoAdapter := v1.NewAdapter()
+
+	cliConfig, err := readConfig()
+	if err != nil {
+		fmt.Println(err)
```
StewartJingga, line 21: is this expected to proceed even when getting error of config reading?

cmd/config.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var (
+	DefaultHost   = "localhost"
+	FileName      = ".guardian"
+	FileExtension = "yaml"
+)
+
+type config struct {
+	Host string `yaml:"host"`
```
StewartJingga, line 18: what's the difference between this config and app.Config?
does it make sense to merge them?

cmd/policies.go
```diff
@@ -0,0 +1,197 @@
...
+		Use:   "update",
+		Short: "update policy",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			b, err := ioutil.ReadFile(filePath)
+			if err != nil {
+				return err
+			}
+
+			var policy domain.Policy
+			switch filepath.Ext(filePath) {
```
StewartJingga, line 146: we could extract this function so `createPolicyCommand` can use it too.

cmd/providers.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return cmd
+}
+
+func listProvidersCommand(c *config) *cobra.Command {
+	return &cobra.Command{
+		Use:   "list",
+		Short: "list providers",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			dialTimeoutCtx, dialCancel := context.WithTimeout(context.Background(), time.Second*2)
```
StewartJingga, line 38: I see these connection codes being used frequently, can we wrap it to a function?
e.g.
```
ctx, cancel := connectToHost(ctx)
```

cmd/config.go
rahmatrhd: I was also having the same doubt, but the difference for cmd.config is that it is only the config for the CLI to run, while app.Config is more into the server's config

cmd/providers.go
rahmatrhd: it's a tricky one, we're deferring `dialCancel()` and `conn.Close()` from here only. I tried to put them into a function`createClient()` that also returns a function that calls both `dialCancel()` and `conn.Close()` for cancellation but it's not possible

StewartJingga (approved):
LGTM

PR #36, opened 2021-08-16
feat: send notification for expiring access reminder

rahmatrhd:


appeal/job_handler.go
```diff
@@ -57,3 +65,40 @@ func (h *JobHandler) RevokeExpiredAccess() error {
 	log.Println(string(result))
 	return nil
 }
+
+func (h *JobHandler) NotifyAboutToExpireAccess() error {
+	daysBeforeExpired := []int{7, 3, 1}
```
rahmatrhd, line 70: will make this configurable in the upcoming PR together with making the notification messages configurable

appeal/job_handler.go
```diff
@@ -57,3 +65,40 @@ func (h *JobHandler) RevokeExpiredAccess() error {
...
+
+		filters := map[string]interface{}{
+			"statuses":           []string{domain.AppealStatusActive},
+			"expiration_date_gt": from,
+			"expiration_date_lt": to,
+		}
+
+		appeals, err := h.appealService.Find(filters)
+		if err != nil {
+			h.logger.Error(err.Error())
```
StewartJingga, line 85: should we add more information such as `dayBeforeExpired` (e.g. 7, 3, or 1)?

also if it fails, would it be better to `continue` or else the remaining actions will be run.
e.g. 
```
if err != nil {
	h.logger.Errorf("error finding appeals \"%d\" days before expiration: %s", err.Error())
        continue
}
```

app/server.go
```diff
@@ -103,6 +103,10 @@ func RunServer(c *Config) error {
 			CronTab: "0/20 * * * *",
 			Func:    appealJobHandler.RevokeExpiredAccess,
 		},
+		{
+			CronTab: "0 9 * * *", // at 09.00
```
StewartJingga, line 107: will this configurable as well in the next update?

rahmatrhd: yes

rahmatrhd:
@StewartJingga updated

StewartJingga (approved):
LGTM

PR #37, opened 2021-08-17
feat: add more filters and payload fields in some endpoints

rahmatrhd:


approval/repository.go
```diff
@@ -14,9 +21,22 @@ func NewRepository(db *gorm.DB) *repository {
 	return &repository{db}
 }
 
-func (r *repository) GetPendingApprovals(approverEmail string) ([]*domain.Approval, error) {
+func (r *repository) ListApprovals(filters map[string]interface{}) ([]*domain.Approval, error) {
```
StewartJingga, line 24: was thinking if it is better if `ListApprovals` is accepting `ListFilter` struct as the parameter.
```
func (r *repository) ListApprovals(filter ListFilter) ([]*domain.Approval, error)

// using it
repo.ListApprovals(repo.ListFilter{
  User: req.User(),
  Statuses: req.GetStatuses(),
})
```

wdyt?

api/handler/v1/grpc.go
```diff
@@ -418,8 +418,9 @@ func (s *GRPCServer) RevokeAppeal(ctx context.Context, req *pb.RevokeAppealReque
 	if err != nil {
 		return nil, status.Error(codes.Internal, "failed to get metadata: actor")
 	}
+	revoke := req.GetReason().GetReason()
```
StewartJingga, line 421: should we rename this to `reason` instead?

StewartJingga (approved):
LGTM

PR #38, opened 2021-08-24
feat: make notification messages configurable

rahmatrhd:


notifier/client.go
```diff
@@ -0,0 +1,58 @@
...
+		result = templates.AppealApproved
+	case domain.NotificationTypeAppealRejected:
+		result = templates.AppealRejected
+	case domain.NotificationTypeApproverNotification:
+		result = templates.ApproverNotification
+	case domain.NotificationTypeExpirationReminder:
+		result = templates.ExpirationReminder
+	}
+
+	result = strings.Replace(result, "{{resource_name}}", message.Variables.ResourceName, -1)
```
AkarshSatija, line 51: Let's utilise go template instead of string replace?

domain/notifier.go
```diff
@@ -1,10 +1,41 @@
...
+
+const (
+	NotificationTypeExpirationReminder   = "ExpirationReminder"
+	NotificationTypeAppealApproved       = "AppealApproved"
+	NotificationTypeAppealRejected       = "AppealRejected"
+	NotificationTypeAccessRevoked        = "AccessRevoked"
+	NotificationTypeApproverNotification = "ApproverNotification"
+)
+
+type NotificationVariables struct {
```
AkarshSatija, line 25: It'd be difficult to scale variables in a struct, should instead use a map interface and not have it here.
Parser method can validate if any of the variables is missing from the body when called.

notifier/client.go
rahmatrhd: agree on this, implemented the changes 👍

AkarshSatija (approved):

PR #41, opened 2021-08-26
feat: add gcloud IAM provider

rahmatrhd:


api/handler/v1/adapter.go
```diff
@@ -154,6 +144,24 @@ func (a *adapter) ToProviderConfigProto(pc *domain.ProviderConfig) (*pb.Provider
 	}, nil
 }
 
+func (a *adapter) ToRole(role *domain.Role) (*pb.Role, error) {
+	permissions := []*structpb.Value{}
```
StewartJingga, line 148: a minor suggestion. If permission is always an object/map, we could use `[]*structpb.Struct{}` to be more explicit.
`structpb.Value` is dynamic and can also hold other types like `string` and `int`.

But changing it requires changing the proto again, so you can decide if the benefits outweigh the complexity.

StewartJingga (approved):
Left a minor suggestion, aside from that LGTM.

api/handler/v1/adapter.go
rahmatrhd: permission is not always expected to be a map, it should be an any (`interface{}`) type. Most of the providers are now using map but we are changing bigquery's permission type to `string` since we have gcloud_iam extracted out as a single provider now

StewartJingga: ahh makes sense then 👌

PR #43, opened 2021-08-28
refactor(logger): migrate logging to use `odpf/salt/log`

### LINKED ISSUE #39: Migrate logging to use odpf/salt
Use salt logger for logging in Guardian. 

https://github.com/odpf/salt/log

wisdommatt:
Signed-off-by: Wisdom Matthew <talk2wisdommatt@gmail.com>

closes #39 

I ran all the tests myself on my machine.


appeal/job_handler.go
```diff
@@ -7,16 +7,16 @@ import (
 	"time"
 
 	"github.com/odpf/guardian/domain"
-	"go.uber.org/zap"
+	saltLog "github.com/odpf/salt/log"
```
ravisuhag, line 10: @wisdommatt Let's avoid alias here as well.

logger/logger.go
```diff
@@ -1,38 +1,14 @@
 package logger
```
ravisuhag, line 1: @wisdommatt We can remove this entire file now and move init of salt loger to server file only.

appeal/job_handler.go
wisdommatt: The reason for the alias is because that particular file imports the `log` std package

What should i do in this situation ?

logger/logger.go
wisdommatt: Okay, i will do that now

appeal/job_handler.go
rahmatrhd: @wisdommatt I think we can replace those logs from `log` package with the salt's one

wisdommatt: > @wisdommatt I think we can replace those logs from `log` package with the salt's one

Noted

wisdommatt: I just pushed a commit to resolve this

ravisuhag:
@wisdommatt Log level can now move to server config. 
https://github.com/odpf/guardian/blob/main/app/server.go#L42

logger/logger.go
ravisuhag: Log level can now move to https://github.com/odpf/guardian/blob/main/app/server.go#L42

wisdommatt: > Log level can now move to https://github.com/odpf/guardian/blob/main/app/server.go#L42

I just pushed a commit that resolves this

rahmatrhd (changes requested):
can you help to update the `config.yaml.example` to match with the current service config as well?

Apart from that, the code LGTM

wisdommatt:
> can you help to update the `config.yaml.example` to match with the current service config as well?
> 
> Apart from that, the code LGTM

The only field in `ServiceConfig` struct that is not in `config.yaml.example` is the `IAM` field

ravisuhag:
@wisdommatt I think you closed this by mistake?

rahmatrhd:
@wisdommatt ah my bad, I updated it in this PR only https://github.com/odpf/guardian/pull/38. I'll update mine from there, you can ignore it. Sorry.

rahmatrhd (approved):

wisdommatt:
> @wisdommatt I think you closed this by mistake?

Yeah, but i just re-opened it

@ravisuhag you can now merge the PR

PR #46, opened 2021-09-03
feat: added appeal validation

ishanarya0:
add ValidateAppeal(a *domain.Appeal) method in provider service to validate the following things on appeal creation:

- [x] appeal options (access durations)
- [x] selected role

For BigQuery Provider:
- [x] validate permissions if they are valid/exist on bigquery provider creation


provider/metabase/config.go
```diff
@@ -49,9 +49,7 @@ func (c *Credentials) Decrypt(decryptor domain.Decryptor) error {
 	return nil
 }
 
-type PermissionConfig struct {
-	Name string `json:"name" mapstructure:"name" validate:"required"`
-}
+type PermissionConfig string
```
rahmatrhd, line 52:
```suggestion
type Permission string
```
Does this variable name sound better to you?

provider/service.go
```diff
@@ -125,6 +125,40 @@ func (s *Service) GetRoles(id uint, resourceType string) ([]*domain.Role, error)
...
+	provider := s.getProvider(p.Type)
+	roles, err := provider.GetRoles(p.Config, resourceType)
+	if err != nil {
+		return err
+	}
+
+	isRoleExists := false
+	for _, role := range roles {
+		if a.Role == role.Name {
+			isRoleExists = true
```
rahmatrhd, line 143:
```suggestion
			isRoleExists = true
			break
```

ishanarya0: Added the break, thanks!

provider/metabase/config.go
ishanarya0: Yes, LGTM but if we keep PermissionConfig, then it would be consistent with all other providers. WDYT @rahmatrhd ?

rahmatrhd: yes, let's rename all of them then if it is better

ishanarya0: Sure! Done with the changes.

rahmatrhd (approved):

PR #52, opened 2021-09-23
chore: ignore .vscode config folder

rahmatrhd:


ravisuhag:
@rahmatrhd I would suggest avoiding putting it in VCS. We should leave this up to devs based on which editor they use.

rahmatrhd:
@ravisuhag understood that this is vscode specific, was thinking to share this "recommended configuration" for devs who use vscode, otherwise, they can just ignore it. That's why I only allow `launch.json` file in the .gitignore. What do you think?

StewartJingga:
I'd have to agree with @ravisuhag on this, this is way too specific and the value gained is not that significant.
And you will need keep checking and sync the `launch.json` to make sure it is still relevant for any updates.

rahmatrhd:
Understood, will ignore the whole .vscode then

PR #53, opened 2021-09-24
feat: update format and guide on policy cmd

ravisuhag:


cmd/policy.go
```diff
@@ -4,33 +4,53 @@ import (
...
 	"github.com/spf13/cobra"
 )
 
-func policiesCommand(c *app.CLIConfig, adapter v1.ProtoAdapter) *cobra.Command {
+//  PolicyCmd is the root command for the policies subcommand.
+func PolicyCmd(c *app.CLIConfig, adapter v1.ProtoAdapter) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "policies",
-		Short: "manage policies",
+		Use:   "policy",
```
rahmatrhd, line 21: any specific reason for changing the command to singular word?

cmd/resources.go
```diff
@@ -41,34 +56,45 @@ func listResourcesCommand(c *app.CLIConfig) *cobra.Command {
 				return err
 			}
 
-			t := getTablePrinter(os.Stdout, []string{"ID", "PROVIDER", "TYPE", "URN", "NAME"})
-			for _, r := range res.GetResources() {
-				t.Append([]string{
+			report := [][]string{}
+
+			resources := res.GetResources()
+			fmt.Printf(" \nShowing %d of %d policies\n \n", len(resources), len(resources))
```
rahmatrhd, line 62:
```suggestion
			fmt.Printf(" \nShowing %d of %d resources\n \n", len(resources), len(resources))
```

cmd/resources.go
```diff
@@ -6,28 +6,43 @@ import (
...
-func resourcesCommand(c *app.CLIConfig) *cobra.Command {
+func ResourceCmd(c *app.CLIConfig) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "resources",
-		Short: "manage resources",
+		Use:   "resource",
+		Short: "Manage resources",
+		Example: heredoc.Doc(`
+			$ guardian resource list
+			$ guardian resource metadata --id=1
```
rahmatrhd, line 23:
```suggestion
			$ guardian resource metadata set --id=1 key=value
```

cmd/policy.go
ravisuhag: Makes it easier to use. Checked some other good clis for reference as well. e.g. github etc.

ravisuhag: @rahmatrhd What do you think?

rahmatrhd: Yeah, as we discussed, we will use singular word and add the plural as the alias

rahmatrhd (approved):

PR #54, opened 2021-09-24
feat: soft delete a resource when deleted from provider

ishanarya0:
Requirements: 

- [x] providerService: introduce one field to flag if the resource is deleted or not 
    - is_deleted: boolean  
    - update is_deleted => true when collecting resources from provider 

- [x] GET /resources
    - in resourceService.Find() when calling resourceRepository.Find() add filter is_deleted=false

- [x] appeal creation:
   - validate if the designated resource is deleted or not


ishanarya0:
Added a filter to List Resource as: /resources/:is_deleted

When `is_deleted` is set `false` -> Guardian will return only non-deleted resources.
When `is_deleted` is set `true` -> Guardian will return all the resources.

ishanarya0:
If a resource for which appeal in being created is already deleted, Guardian will return an error message.

resource/repository.go
```diff
@@ -37,6 +38,11 @@ func (r *Repository) Find(filters map[string]interface{}) ([]*domain.Resource, e
 	if conditions.IDs != nil {
 		db = db.Where(conditions.IDs)
 	}
+
+	if filters["isDeleted"] != nil {
```
ishanarya0, line 42: @rahmatrhd Added this to first validate if a `isDeleted ` filter is being used or not. Using conditions.IsDeleted would be misleading as it would be set to false when the filter is not being applied.

api/handler/v1/grpc.go
```diff
@@ -226,7 +226,10 @@ func (s *GRPCServer) UpdatePolicy(ctx context.Context, req *pb.UpdatePolicyReque
 }
 
 func (s *GRPCServer) ListResources(ctx context.Context, req *pb.ListResourcesRequest) (*pb.ListResourcesResponse, error) {
-	resources, err := s.resourceService.Find(map[string]interface{}{})
+	isDeleted := req.GetIsDeleted()
+	resources, err := s.resourceService.Find(map[string]interface{}{
+		"isDeleted": isDeleted,
```
rahmatrhd, line 231:
```suggestion
		"is_deleted": isDeleted,
```

rahmatrhd (approved):

PR #61, opened 2021-10-04
feat: add filters to list resource api

ishanarya0:
Add following filters in list resources api

- [ ] provider_type  
- [ ] provider_urn  
- [ ] type  
- [ ] urn  
- [ ] name  
- [ ]  any other fields will get applied to the resource.details


ishanarya0:
To fetch deleted resources, set is_deleted param to true, else it won't return deleted resource in any other cases.

PR #82, opened 2021-11-02
feat(policy): read iam config from policy 

### LINKED ISSUE #70: Identity manager config at the policy level
1. Move identity manager config from app config to the policy
https://github.com/odpf/guardian/blob/99ef4e3c06d1308ead17f9a9d9175bffe10a1631/iam/client.go#L14-L22
2. Read identity manager config from policy

rahmatrhd:
closes #70

- add new field for IAM config in policy
  - `IAM map[string]interface{}`
 - remove iam config in app config


domain/iam.go
```diff
@@ -1,5 +1,17 @@
...
+type IAMProviderType string
+
+const (
+	IAMProviderTypeShield IAMProviderType = "shield"
+	IAMProviderTypeHTTP   IAMProviderType = "http"
+)
+
+type IAMConfig struct {
+	Provider IAMProviderType `json:"provider" yaml:"provider" validate:"required,oneof=http shield"`
+	Config   interface{}     `json:"config" yaml:"config" validate:"required"`
```
StewartJingga, line 12: will `map[string]interface{}` be clearer here? or it could be in other format as well?

rahmatrhd: I tried using `map[string]interface{}` type instead of `interface{}` and turns out it added more complexity even though it's correct that the value of the config should always be a map/object. I'll give you an example scenario here:

Creating a new policy:
1. new policy input is in json/yaml and the handler parses it to `domain.Policy` type
```yaml
# policy config yaml
...
iam:
  provider: http
  config:
    url: ...
    auth: ...
```
2. at this point `Policy.IAMConfig.Config` still on `map[string]interface{}` type
3. in policyService, iamManager parses the `IAMConfig.Config` to `domain.SensitiveConfig` interface (for now either `iam.ShieldClientConfig{}` or `iam.HTTPClientConfig{}`)
4. and store the parsed iam config struct back to the `Policy.IAMConfig.Config`
5. policy service then encrypt and store the config

Step 4. is possible because the type of `Policy.IAMConfig.Config` is `interface{}`. Type `map[string]interface{}` makes us unable to store `domain.SensitiveConfig` interface which is needed by policy service before storing to db

StewartJingga (approved):
LGTM

Issue #62, opened 2021-10-05
Guardian fetches all resource types regardless the config


rahmatrhd:
# Context

Once a provider is registered, guardian will fetch **all** the supported resource types from the provider, even though the provider configuration only specifying for some resource types.

Example:
```yaml
# provider config
type: bigquery
resources:
- type: dataset
  policy: ...
  roles: ...
  ...
...
```
With this config, guardian not only will fetch all datasets, but also all tables.
This could cause an issue if a user is creating an appeal to a table resource from that provider because the policy and roles are not configured yet for that table resource type.

# Proposed solution

Make guardian to only fetch the specified resource types based on the provider config.

PR #85, opened 2021-11-12
feat(provider): only fetch resource types that are added in provider config

rahmatrhd:
Closes https://github.com/odpf/guardian/issues/62

Changes:
1. trigger fetch resources once provider created
2. fetch only registered resource types


StewartJingga (approved):
LGTM

PR #91, opened 2021-12-13
fix: parse `allowed_account_types` in provider config

### LINKED ISSUE #89: `allowed_account_types` field in provider config is ignored
**Describe the bug**
1. on provider create/update, `allowed_account_types` field is ignored and not being passed to db
2. empty `allowed_account_types` failed to use default value automatically

rahmatrhd:
closes #89 

proto changes: https://github.com/odpf/proton/pull/74


appeal/service.go
```diff
@@ -129,25 +129,27 @@ func (s *Service) Create(appeals []*domain.Appeal) error {
 			activeAppeals[a.AccountID][a.ResourceID] != nil &&
 			activeAppeals[a.AccountID][a.ResourceID][a.Role] != nil {
 
-			if p.Config.Appeal.AllowActiveAccessExtensionIn == "" {
-				return ErrAppealFoundActiveAccess
-			}
+			if p.Config.Appeal != nil {
```
snorlacs, line 132: @rahmatrhd what is the expected behaiour if `p.Config.Appeal` is nil

snorlacs (approved):

appeal/service.go
rahmatrhd: guardian will ignore the appeal config, because rules like `AllowActiveAccessExtensionIn` and `AllowPermanentAccess ` are not present

PR #93, opened 2021-12-14
feat: introduce approval reason

rahmatrhd:


approval/repository_test.go
```diff
@@ -57,7 +57,7 @@ func (s *RepositoryTestSuite) TearDownTest() {
 }
 
 func (s *RepositoryTestSuite) TestBulkInsert() {
-	expectedQuery := regexp.QuoteMeta(`INSERT INTO "approvals" ("name","index","appeal_id","status","actor","policy_id","policy_version","created_at","updated_at","deleted_at") VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10),($11,$12,$13,$14,$15,$16,$17,$18,$19,$20) RETURNING "id"`)
+	expectedQuery := regexp.QuoteMeta(`NSERT INTO "approvals" ("name","index","appeal_id","status","actor","reason","policy_id","policy_version","created_at","updated_at","deleted_at") VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11),($12,$13,$14,$15,$16,$17,$18,$19,$20,$21,$22) RETURNING "id"`)
```
StewartJingga, line 60: typo on `NSERT`, should be `INSERT` right?

StewartJingga: also why it did not break the test? 🤔

StewartJingga (changes requested):
Found a typo and some bugs probably in the test because the typo does not break the test.

approval/repository_test.go
rahmatrhd: Maybe because it is regex, will check

rahmatrhd: fixed, `NSERT` was working because it is a regex checking and it still fulfills

StewartJingga (approved):
LGTM

PR #95, opened 2021-12-16
chore: apply default sort by status for appeals and approvals

rahmatrhd:


approval/repository.go
```diff
@@ -53,13 +56,38 @@ func (r *repository) ListApprovals(conditions *domain.ListApprovalsFilter) ([]*d
...
 
-	db = db.Clauses(clause.OrderBy{
-		Expression: clause.Expr{
-			SQL:                `ARRAY_POSITION(ARRAY[?], "approvals"."status"), "updated_at" desc`,
-			Vars:               []interface{}{ApprovalStatusDefaultSort},
-			WithoutParentheses: true,
-		},
-	})
+	if conditions.OrderBy != nil {
+		var orderByClauses []string
```
StewartJingga, line 60: Should we move this code inside the if block to another function for readibility?

StewartJingga (changes requested):

approval/repository.go
rahmatrhd: updated

StewartJingga (approved):

PR #104, opened 2022-01-06
refactor: restructure packages

rahmatrhd:
- [x] move `model` into `store`
- [x] move repositories into `store/postgres`
- [x] move provider clients into `plugins/providers`
- [x] move notifier clients into `plugins/notifiers`
- [x] move iam clients into `plugins/identities`


ravisuhag (approved):
LGTM

PR #102, opened 2022-01-05
feat(cli): improve appeal help messages

ravisuhag:
- Fix usage guide on the wrong command. 
- Improve appeal help messages.

PR #116, opened 2022-01-17
fix: fix bugs

rahmatrhd:
- fix `view resource` command default format
- rename `expression` field to `when` in yaml and json format
- fix requirement `on` (trigger) and `resource` (identifier) fields validation

### DIFF
--- a/cmd/resource.go
+++ b/cmd/resource.go
@@ -78,7 +78,7 @@ func listResourcesCmd(c *app.CLIConfig, adapter handlerv1beta1.ProtoAdapter) *co
 				return err
 			}
 
-			format := cmd.Flag("format").Value.String()
+			format := cmd.Flag("output").Value.String()
 			if format != "" {
 				var resources []*domain.Resource
 				for _, r := range res.GetResources() {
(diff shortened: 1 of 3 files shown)


cmd/resource.go
```diff
@@ -78,7 +78,7 @@ func listResourcesCmd(c *app.CLIConfig, adapter handlerv1beta1.ProtoAdapter) *co
 				return err
 			}
 
-			format := cmd.Flag("format").Value.String()
+			format := cmd.Flag("output").Value.String()
```
ravisuhag, line 81: @rahmatrhd does this denote the format of output? If yes, any reason to call it output and not format?

rahmatrhd: I thought we are moving to this fashion `guardian appeals list -o json`. Few days ago we change it to `output` in some places >> https://github.com/odpf/guardian/pull/102/files#diff-16ff6f0fda308d588b6048ff5bf2a0a208113dadc5b190b36497f90178347f49R158. This is one where we forgot to replace

StewartJingga: IMHO in general `output` sounds less ambiguous than `format`.
Format could mean other things I guess 🤔 if no context provided.

StewartJingga (approved):

core/policy/service_test.go
```diff
@@ -425,6 +425,116 @@ func (s *ServiceTestSuite) TestPolicyRequirements() {
...
+								},
+								Role: "viewer",
+							},
+						},
+					},
+				},
+			},
+		}
+
+		for _, tc := range testCases {
```
StewartJingga, line 494: should each test cases has it's own `s.Run`?
CMIIW but right now `tc.name` is not being used anywhere.

StewartJingga (changes requested):

cmd/resource.go
ravisuhag: Yes, my bad. I changed it to output only for the same reason. :)

core/policy/service_test.go
rahmatrhd: @StewartJingga updated

PR #117, opened 2022-01-18
feat: cli-ux-enhancements

ishanarya0:
Proposed enhancements: 

- [x] Add init command for a policy and provider
- [x] Print better blank state errors - no resource found.
- [x] Show spinner when fetching results in commands (with tty)


ishanarya0:
@ravisuhag [TTY](https://github.com/piotrmurach/tty-spinner) was for ruby applications? Used this [spinner](https://github.com/briandowns/spinner) alternately.

ravisuhag:
> @ravisuhag [TTY](https://github.com/piotrmurach/tty-spinner) is for ruby applications? Used this [spinner](https://github.com/briandowns/spinner) alternately.
@ishanarya0 
With TTY I meant terminal. I have abstracted the spinner with this check-in salt. https://github.com/odpf/salt/blob/main/term/spinner.go
We can use that spinner.

ishanarya0:
> > @ravisuhag [TTY](https://github.com/piotrmurach/tty-spinner) is for ruby applications? Used this [spinner](https://github.com/briandowns/spinner) alternately.
> 
> @ishanarya0 With TTY I meant terminal. I have abstracted the spinner with this check-in salt. https://github.com/odpf/salt/blob/main/term/spinner.go We can use that spinner. When we are using salt spinner, I think we can avoid defer and whitespace in labels as well.

Oh, I'll look into salt's spinner.

ishanarya0:
> > @ravisuhag [TTY](https://github.com/piotrmurach/tty-spinner) is for ruby applications? Used this [spinner](https://github.com/briandowns/spinner) alternately.
> 
> @ishanarya0 With TTY I meant terminal. I have abstracted the spinner with this check-in salt. https://github.com/odpf/salt/blob/main/term/spinner.go We can use that spinner. When we are using salt spinner, I think we can avoid defer and whitespace in labels as well.

We need to keep deferring. As if there is an error before we stop the spinner (which we do before printing the output), the spinner remains on the terminal, and the error message is shown.
 
 Without defer:
 ```
guardian appeal list
Fetching appeal list ⠇context deadline exceeded
```

With defer:
```
guardian appeal list
context deadline exceeded
```

File: cmd/appeal.go [hunk 1]
```diff
@@ -53,6 +53,9 @@ func listAppealsCommand(c *app.CLIConfig) *cobra.Command {
 			$ guardian appeal list --role=viewer
 		`),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			s := term.Spin("Fetching appeal list")
+			defer s.Stop()
```
ravisuhag, line 57: @ishanarya0 Do we need this defer if we are closing spin below explicitly anyway.

ishanarya0: Yes. If an error is returned at line 64 or line 74, then we won't be able to close the spinner on line 80.

ishanarya0: The spinner label and frozen spinner will remain on the terminal, followed by the error message. In order to remove it, we need to close the spinner. Or, if we are okay with this state, then we can remove defer.

Example: https://github.com/odpf/guardian/pull/117#issuecomment-1016250527

Issue #120, opened 2022-01-19
bug: client not configured error is printed during the guardian config init cmd execution


bsushmith:
**Describe the bug**

"client not configured." error is printed even after the config init step.

``` shell
❯ ./guardian config init
config created: /Users/sushmithbhatta/.config/odpf/guardian.yml
client not configured. try running `guardian config init`
```

**To Reproduce**
1. Remove `~/.config/odpf/guardian.yml` file if exists in local
2. Run `guardian config init` after building.
3. See error in cli

**Expected behavior**

No error should be printed during the config init step if it runs successfully.
```
❯ ./guardian config init
config created: /Users/sushmithbhatta/.config/odpf/guardian.yml
```

**Additional context**
issue caused by `defer` - https://github.com/odpf/guardian/blob/main/main.go#L25

PR #139, opened 2022-03-04
fix(cmd): load `host` from flag for core commands

rahmatrhd:
closes: #120 

changes:
- added `host` flag for core commands
- load config from file first, if it's not available, read config fields from flags
- refactor: move `app.CLIConfig` struct to `cmd.Config`

any config stored in the file should be override-able via flag too. Currently, there is only `host`


rahmatrhd:
@ravisuhag pushed new updates. Please help to review

PR #141, opened 2022-03-08
feat: add match, split, contains function in expression evaluation

singhvikash11:
## What?

> Add string function split to [expr](https://github.com/antonmedv/expr) evaluator.

## Why?

> This function will help to create an array using string split function
`{
            "name": "project_regex", "strategy": "auto", "approve_if": "len(Split($appeal.account_id, ',')) > 1"
        }`

### DIFF
--- a/internal/evaluator/expression.go
+++ b/internal/evaluator/expression.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


rahmatrhd:
regex string matching already supported by the expr library. You can use it like `$appeal.account_id matches "/regex/"` as per the [docs](https://github.com/antonmedv/expr/blob/master/docs/Language-Definition.md#string-operators)

rahmatrhd (approved):

PR #145, opened 2022-03-22
feat: add console host url as variable in notification

singhvikash11:


rahmatrhd:
@singhvikash11 if you only want to add the host inside the message we can override the default message directly from guardian config. Example:

```yaml
notifier:
  messages:
    appeal_approved: "Your appeal to {{.resource_name}} with role {{.role}} has been approved, click this link to view https://example.com/dataaccess/requests."
```
or through env variable
```env
NOTIFIER_MESSAGES_APPEAL_APPROVED="Your appeal to {{.resource_name}} with role {{.role}} has been approved, click this link to view https://example.com/dataaccess/requests."
```

code changes should only for adding unavailable variables

singhvikash11:
@rahmatrhd we want to make it conditional based on integration and prod environment so need this variable passed out from config

PR #147, opened 2022-03-23
feat: add curl and openssl to image

bsushmith:
* Add curl and open ssl
* Get certificates from authority for openIDM
* Decode cert file to .pem file


ravisuhag:
@bsushmith Can you add reasons for this change in description.

snorlacs (approved):

rahmatrhd (approved):

PR #155, opened 2022-04-06
feat: audit logging

rahmatrhd:


audit/audit.go
```diff
@@ -0,0 +1,64 @@
+package audit
```
rahmatrhd, line 1: this package can be extracted to salt later

core/appeal/service.go
```diff
@@ -1,6 +1,16 @@
+//go:generate mockery --name=repository --exported
```
mabdh, line 1: We can use `--with-expecter` here so we have more strongly typed mock functions https://lens.golabs.io/graph/d/nWve-U2Gz/elasticsearch-global-search-platform?orgId=1&refresh=30s&var-datasource=tardis&var-environment=production&var-cluster=global-search-production&var-host=All

app/helper.go
```diff
@@ -0,0 +1,30 @@
...
+
+type helper struct {
+	authenticatedUserHeaderKey string
+}
+
+func NewHelper(authenticatedUserHeaderKey string) *helper {
+	return &helper{authenticatedUserHeaderKey}
+}
+
+func (h helper) GetAuthenticatedUser(ctx context.Context) (string, error) {
```
mabdh, line 18: Is it possible to move this to the `user` domain context instead of having a generic `helper` ?

core/appeal/service.go
```diff
@@ -15,72 +25,105 @@ import (
 	"github.com/odpf/salt/log"
 )
 
+var (
```
mabdh, line 28: Can we use `const` here instead?

core/appeal/service.go
```diff
@@ -15,72 +25,105 @@ import (
...
-	appealRepository store.AppealRepository,
-	approvalService approvalService,
-	resourceService resourceService,
-	providerService providerService,
-	policyService policyService,
-	iam domain.IAMManager,
-	notifier notifiers.Client,
-	logger log.Logger,
-) *Service {
+func NewService(opts ServiceOptions) *Service {
```
mabdh, line 108: If we are using options here, should we use it like [golang options](https://www.google.com/search?q=golang+function+options&oq=golang+function+options&aqs=chrome..69i57.2961j0j7&sourceid=chrome&ie=UTF-8) behave? The fields are optional and has default value if not passed.

core/appeal/service.go
```diff
@@ -217,6 +260,10 @@ func (s *Service) Create(appeals []*domain.Appeal) error {
 		return fmt.Errorf("inserting appeals into db: %w", err)
 	}
 
+	if err := s.auditLogger.Log(ctx, AuditKeyBulkInsert, appeals); err != nil {
+		s.logger.Error(fmt.Sprintf("failed to record audit log: %s", err))
```
mabdh, line 264: with `salt/log` we can do like this instead right?
```suggestion
		s.logger.Error("failed to record audit log","err", err)
```

core/appeal/service.go
```diff
@@ -353,15 +400,23 @@ func (s *Service) MakeAction(approvalAction domain.ApprovalAction) (*domain.Appe
 				}
 			}
 
+			auditKey := AuditKeyApprove
+			if approvalAction.Action == "revoke" {
+				auditKey = AuditKeyReject
+			}
+			if err := s.auditLogger.Log(ctx, auditKey, approvalAction); err != nil {
+				s.logger.Error(fmt.Sprintf("failed to record audit log: %s", err))
```
mabdh, line 408: same with this logging

api/handler/v1beta1/grpc.go
```diff
@@ -145,7 +146,13 @@ func (s *GRPCServer) CreateProvider(ctx context.Context, req *guardianv1beta1.Cr
 		URN:    providerConfig.URN,
 		Config: providerConfig,
 	}
-	if err := s.providerService.Create(p); err != nil {
+
+	user, err := s.getUser(ctx)
+	if err != nil {
+		return nil, err
```
mabdh, line 152: any reason why we don't define error status code here and return this error directly instead?

api/handler/v1beta1/grpc.go
```diff
@@ -172,7 +179,13 @@ func (s *GRPCServer) UpdateProvider(ctx context.Context, req *guardianv1beta1.Up
 		URN:    providerConfig.URN,
 		Config: providerConfig,
 	}
-	if err := s.providerService.Update(p); err != nil {
+
+	user, err := s.getUser(ctx)
+	if err != nil {
+		return nil, err
+	}
+	ctx = audit.WithActor(ctx, user)
```
mabdh, line 187: If this is done in every/most of API, should we move it to grpc interceptor?

app/helper.go
rahmatrhd: this is actually no longer being used, will remove this

app/server.go
```diff
@@ -86,32 +88,61 @@ func RunServer(c *Config) error {
...
-		providerService,
-		iamManager,
+	auditRepository := audit_repos.NewPostgresRepository(db)
+	auditLogger := audit.New(
+		audit.WithRepository(auditRepository),
+		audit.WithAppDetails(audit.AppDetails{
+			Name:    "guardian",
+			Version: Version,
+		}),
+		audit.WithTrackIDExtractor(func(ctx context.Context) string {
```
mabdh, line 98: TraceID or TrackID ?

app/server.go
```diff
@@ -86,32 +88,61 @@ func RunServer(c *Config) error {
...
+			if md, ok := metadata.FromIncomingContext(ctx); ok {
+				if rawTraceID := md.Get(c.AuditLogTraceIDHeaderKey); len(rawTraceID) > 0 {
+					return rawTraceID[0]
+				}
+			}
+			return ""
```
mabdh, line 104: Does it mean if TraceID not found in the header, we still audit it but with empty TraceID?

app/server.go
```diff
@@ -124,15 +155,15 @@ func RunServer(c *Config) error {
 	tasks := []*scheduler.Task{
 		{
 			CronTab: c.Jobs.FetchResourcesInterval,
-			Func:    jobHandler.FetchResources,
+			Func:    func() error { return jobHandler.FetchResources(context.Background()) },
```
mabdh, line 158: Should we use context `WithTimeout` or `WithDeadline` to make sure job is not stucked?

app/server.go
rahmatrhd: should be traceID

app/server.go
```diff
@@ -236,6 +267,13 @@ func Migrate(c *Config) error {
 	if err != nil {
 		return err
 	}
+
+	db := store.DB()
+	auditRepository := audit_repos.NewPostgresRepository(db)
```
mabdh, line 272: can just simply do like this?
```suggestion
	auditRepository := audit_repos.NewPostgresRepository(store.DB())
```

app/server.go
```diff
@@ -265,6 +303,8 @@ func makeHeaderMatcher(c *Config) func(key string) (string, bool) {
 		switch strings.ToLower(key) {
 		case strings.ToLower(c.AuthenticatedUserHeaderKey):
 			return key, true
+		case "x-trace-id": // TODO: make key configurable
```
mabdh, line 306: Is it intentionally will be done later or this is just missing in this PR?

rahmatrhd: fixing it

core/appeal/service_test.go
```diff
@@ -131,33 +137,33 @@ func (s *ServiceTestSuite) TestCreate() {
 	}
 	s.Run("should return error if got error from resource service", func() {
 		expectedError := errors.New("resource service error")
-		s.mockResourceService.On("Find", mock.Anything).Return(nil, expectedError).Once()
+		s.mockResourceService.On("Find", mock.Anything, mock.Anything).Return(nil, expectedError).Once()
```
mabdh, line 140: when we are using `mocks` it is better if we don't use `mock.Anything` if it is possible. Also at the end, we could also try asserting all expected mocks with `mock.AssertExpectations(t)`

core/provider/service.go
```diff
@@ -14,51 +20,75 @@ import (
 	"github.com/odpf/salt/log"
 )
 
+var (
```
mabdh, line 23: same here for `const`

app/server.go
rahmatrhd: we should generate a TraceID. Need to discuss the trace ID convention first. Marking it as TODO

api/handler/v1beta1/grpc.go
rahmatrhd: that sounds nice, let me check

api/handler/v1beta1/grpc.go
rahmatrhd: in case of `CreateProvider`, authenticated user email is only required for audit logging purpose, before, we're not needing it for the provider creation. And I also just realize we can log.Error it instead of returning the error then ☝️, wdyt? 

There's an example where err from `getUser` returned with the error code accordingly like in `CreateAppeal` because the user email is needed for appeal creation.

jobs/fetch_resources.go
```diff
@@ -1,5 +1,12 @@
...
-func (h *handler) FetchResources() error {
-	return h.providerService.FetchResources()
+import (
+	"context"
+
+	"github.com/odpf/guardian/pkg/audit"
+)
+
+func (h *handler) FetchResources(ctx context.Context) error {
+	ctx = audit.WithActor(ctx, "system")
```
mabdh, line 10: saw this in several places, should we define a new function called `WithSystemActor` ?

core/appeal/service.go
rahmatrhd: just learned that this should've been dependencies instead of options, changing the struct name to `ServiceDeps` will be more reasonable now in my opinion

app/server.go
rahmatrhd: skipping it for now as I'm not sure what the duration/deadline value should be

jobs/fetch_resources.go
rahmatrhd: the current one is still simple enough I would say, it is also just used in two different places

core/appeal/service.go
rahmatrhd: nice, will check if we're gonna need this now

api/handler/v1beta1/grpc.go
rahmatrhd: I created `audit.UnaryServerInterceptor` to avoid this repetition as suggested to use grpc interceptor

api/handler/v1beta1/grpc.go
mabdh: Returning error or not (just logging it) depends on the context though, whether it is required or optional.

But in case it is optional and the worst case would be there is no user information, would it be ok for audit log to have empty 'actor'?

app/server.go
mabdh: timeout could also be passed from config (specified from cli), if not specified, we can make it indefinite. but I think this could be done later on.

core/appeal/service.go
```diff
@@ -349,19 +396,27 @@ func (s *Service) MakeAction(approvalAction domain.ApprovalAction) (*domain.Appe
 			}
 			if len(notifications) > 0 {
 				if err := s.notifier.Notify(notifications); err != nil {
-					s.logger.Error(err.Error())
+					s.logger.Error("failed to send notifications", "error", err.Error())
 				}
 			}
 
+			auditKey := AuditKeyApprove
+			if approvalAction.Action == "revoke" {
```
mabdh, line 404: should we make `const` out of all possible actions and store it in a central place instead of hardcoding it here?

core/provider/service.go
```diff
@@ -69,61 +99,68 @@ func (s *Service) Create(p *domain.Provider) error {
...
 
+	if err := s.auditLogger.Log(ctx, AuditKeyCreate, p); err != nil {
+		s.logger.Error("failed to record audit log", "error", err)
+	}
+
 	go func() {
-		s.logger.Info(fmt.Sprintf("fetching resources for %s", p.URN))
-		resources, err := s.getResources(p)
+		s.logger.Info("fetching resources", "provider_urn", p.URN)
+		resources, err := s.getResources(ctx, p)
```
mabdh, line 116: any reason why we do this inside go routine?
since we are using the context propagated from the parent function, if this goroutine takes longer time and the context is cancelled earlier, these will get cancelled too.
therefore, it is not different whether using goroutine or not.

audit/audit.go
mabdh: Should we add `TODO` in it @rahmatrhd ?

jobs/fetch_resources.go
mabdh: Hmm true but IMO defining it in one place and reuse it in other places would make it more consistent (in case there is a typo) and easier to refactor later on.

core/provider/service.go
rahmatrhd: we want to fire and forget `FetchResources` the moment a provider is created.

I see, wasn't aware of that. But using a new context instance also means losing the audit actor info from the original ctx right?

mabdh: yes but that is fine right? we don't need embedded data in ctx inside the go routine and the logic inside go routines does not need to be depend on the ctx, we can just create a new context with its own timeout and kill go routine if it reaches the timeout.

rahmatrhd: I was thinking we need to keep the actor from the original ctx, but I think creating new context with `"system"` actor should be fine for this fetching resources post provider create 👍

api/handler/v1beta1/grpc.go
rahmatrhd: yes, that's exactly the case, authorization is optional on these methods while we're getting the actor info from the authorized user email header. We are also still not sure how this should be handled.

audit/audit.go
rahmatrhd: created this issue instead https://github.com/odpf/salt/issues/31

pkg/audit/model.go
```diff
@@ -0,0 +1,17 @@
...
+	Version string
+}
+
+type Log struct {
+	TraceID   string
+	Timestamp time.Time
+	Action    string
+	Actor     string
+	App       *AppDetails
+	Data      interface{}
```
rahmatrhd, line 16: add `Metadata interface{}`

rahmatrhd: and put TraceID inside metadata

krtkvrm (approved):

PR #156, opened 2022-04-08
rfc: edit provider configuration to add resource conditionally

singhvikash11:


docs/docs/rfcs/00000000_add_resource_conditionally.md
```diff
@@ -0,0 +1,33 @@
...
+
+Add a scope of condition in provider configuration to filter out on resources.
+
+#### Using expression
+
+Expression can be used to check whether Guardian should add the resources or not.
+
+```yaml
+resources:
+  add_if: $resource.urn endsWith 'keyset'
```
rahmatrhd, line 25: can we rename this field to `filter`?

rahmatrhd (changes requested):

docs/docs/rfcs/00000000_add_resource_conditionally.md
singhvikash11: @rahmatrhd sure, what you suggest for this filter name

ravisuhag: @singhvikash11 `filter` itself.

bsushmith: Wouldn't `add_if` be more appropriate here? `filter` can mean either `filtering out` or `include these only`. `add_if` wouldn't have that confusion.

mabdh: `filter` is fine I guess and can be utilized for generic purpose, or we can use `where` to make it more concise/

rahmatrhd: @bsushmith the `filter` field will have expression value which expected to return boolean, from there we can infer whether the resource will be included or filtered out. Same as javascript's `Array.filter()` api

rahmatrhd:
converted this PR to an issue here: https://github.com/odpf/guardian/issues/313

PR #157, opened 2022-04-08
feat: support metabase v0.42 version and also make it backward compat…

singhvikash11:
…ible

### DIFF
--- a/plugins/providers/metabase/client.go
+++ b/plugins/providers/metabase/client.go
@@ -139,8 +139,10 @@ func (c *client) GetDatabases() ([]*Database, error) {
 	if _, err := c.do(req, &response); err == nil {
 		if v, ok := response.([]interface{}); ok {
 			err = mapstructure.Decode(v, &databases) // this is for metabase v0.37
+		} else if v, ok := response.(map[string]interface{}); ok && v["data"] != nil {
+			err = mapstructure.Decode(v["data"], &databases) // this is for metabase v0.42
 		} else {
-			err = mapstructure.Decode(response.(map[string]interface{})["data"], &databases) // this is for metabase v0.42
+			return databases, ErrInvalidApiResponse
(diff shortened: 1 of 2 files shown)


plugins/providers/metabase/client.go
```diff
@@ -133,11 +135,20 @@ func (c *client) GetDatabases() ([]*Database, error) {
...
 	var databases []*Database
-	if _, err := c.do(req, &databases); err != nil {
-		return nil, err
-	}
+	var response interface{}
+	if _, err := c.do(req, &response); err == nil {
+		if v, ok := response.([]interface{}); ok {
+			err = mapstructure.Decode(v, &databases) // this is for metabase v0.37
+		} else {
+			err = mapstructure.Decode(response.(map[string]interface{})["data"], &databases) // this is for metabase v0.42
```
rahmatrhd, line 143: `response.(map[string]interface{})`
need to handle this type assertion to avoid runtime panic

rahmatrhd (changes requested):

plugins/providers/metabase/client.go
```diff
@@ -133,11 +135,22 @@ func (c *client) GetDatabases() ([]*Database, error) {
 	}
 
 	var databases []*Database
-	if _, err := c.do(req, &databases); err != nil {
-		return nil, err
-	}
+	var response interface{}
+	if _, err := c.do(req, &response); err == nil {
```
rahmatrhd, line 139: let's keep handling error by `err != nil` to keep the code style consistency

rahmatrhd (changes requested):

plugins/providers/metabase/client.go
singhvikash11: check now

rahmatrhd (approved):

Issue #159, opened 2022-04-11
Duplicate active appeal records in guardian table


bsushmith:
**Describe the bug**
Consider an appeal is created which is tied to a policy containing only one auto approval step and the auto approval condition becomes true. If the provider config for that appeal allows permanent access, then the [create method](https://github.com/odpf/guardian/blob/main/core/appeal/service.go#L97) creates another record with an active status into the appeal table.

**To Reproduce**

* Create a policy definition with only one auto approval step and no manual ones.
* Tie up a provider with this policy and enable `allow_permanent_access` in the provider appeal config.
* Create an appeal
* Create an appeal again for the same resource and role with the same details (given that the appeal is within the `allow_active_access_extension_in` duration) 
* There will be two active records in the table instead of only one.

**Expected behavior**
There should only be one active record in the table.

**Additional context**
This is happening because the logic to check if there is an already active appeal and to mark it as terminated is embedded within the [MakeAction](https://github.com/odpf/guardian/blob/main/core/appeal/service.go#L280) function call., which does not get triggered at all in the above case. 

**version**
`latest` on main branch

PR #160, opened 2022-04-12
fix: terminate old active appeal

bsushmith:
Fixes: #159

### DIFF
--- a/core/appeal/service.go
+++ b/core/appeal/service.go
@@ -124,6 +124,7 @@ func (s *Service) Create(appeals []*domain.Appeal) error {
 
 	notifications := []domain.Notification{}
 
+	var oldExtendedAppeals []*domain.Appeal
 	for _, appeal := range appeals {
 		appeal.SetDefaults()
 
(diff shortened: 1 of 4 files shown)


core/appeal/service.go
```diff
@@ -173,6 +173,25 @@ func (s *Service) Create(appeals []*domain.Appeal) error {
...
+					Statuses:   []string{domain.AppealStatusActive},
+				})
+				if err != nil {
+					return fmt.Errorf("unable to retrieve existing active appeal from db: %w", err)
+				}
+
+				if len(activeAppeals) > 0 {
+					oldExtendedAppeal = activeAppeals[0]
+					oldExtendedAppeal.Terminate()
+					if err := s.repo.Update(oldExtendedAppeal); err != nil {
```
rahmatrhd, line 190: let's use BulkUpsert to update existing appeal and insert ones at the same time

rahmatrhd (changes requested):

core/appeal/service.go
bsushmith: Made the suggested changes.

jobs/handler.go
```diff
@@ -9,6 +9,7 @@ import (
 type appealService interface {
 	Find(*domain.ListAppealsFilter) ([]*domain.Appeal, error)
 	Revoke(id, actor, reason string) (*domain.Appeal, error)
+	Create(appeals []*domain.Appeal) error
```
rahmatrhd, line 12: Is this required?

bsushmith: I made this change, as I was thinking of mocking appealService itself for a usecase. Didn't need it later, but left it as it is, since it felt that this interface should have the contract of create appeal also.

rahmatrhd: we're trying to have interface defined in the user part and only define methods that the user's using as per [this docs](https://github.com/golang/go/wiki/CodeReviewComments#interfaces)

rahmatrhd (approved):

PR #162, opened 2022-04-20
feat: adding table, group as metabase resource and add group-database

singhvikash11:


plugins/providers/metabase/resource.go
```diff
@@ -54,6 +76,76 @@ func (d *Database) ToDomain() *domain.Resource {
...
+}
+
+func (g *Group) FromDomain(r *domain.Resource) error {
+	if r.Type != ResourceTypeGroup {
+		return ErrInvalidResourceType
+	}
+
+	groupUrn := strings.Split(r.URN, ":")
+	if len(groupUrn) != 2 {
+		return ErrInvalidDatabaseURN
```
bsushmith, line 120: shouldn't this be `ErrInvalidGroupURN`?

plugins/providers/metabase/resource.go
```diff
@@ -54,6 +76,76 @@ func (d *Database) ToDomain() *domain.Resource {
...
+	}
+	id, err := strconv.Atoi(groupUrn[1])
+	if err != nil {
+		return err
+	}
+
+	g.ID = id
+	g.Name = r.Name
+	g.DatabaseResources = r.Details["database"].([]map[string]interface{})
```
bsushmith, line 129: `map[string]interface{}` is too generic. Can this be made specific?

plugins/providers/metabase/client.go
```diff
@@ -18,6 +19,7 @@ import (
 type MetabaseClient interface {
 	GetDatabases() ([]*Database, error)
 	GetCollections() ([]*Collection, error)
+	GetGroups() ([]*Group, map[string][]map[string]interface{}, map[string][]map[string]interface{}, error)
```
bsushmith, line 22: lets have concrete types for this method.

bsushmith: Can you also name these parameters.

plugins/providers/metabase/client.go
```diff
@@ -170,6 +172,134 @@ func (c *client) GetCollections() ([]*Collection, error) {
...
+		groupMap[fmt.Sprintf("group:%d", group.ID)] = group
+	}
+
+	addResourceToGroup(databaseResourceGroups, groupMap, "database")
+	addResourceToGroup(collectionResourceGroups, groupMap, "collection")
+
+	return groups, databaseResourceGroups, collectionResourceGroups, err
+}
+
+func (c *client) fetchGroups(wg *sync.WaitGroup, groups *[]*Group, err error) {
```
aok1j1, line 203: `*[]*Groups`? - can we have a map here itself like the others.

plugins/providers/metabase/client.go
```diff
@@ -129,7 +131,7 @@ func NewClient(config *ClientConfig) (*client, error) {
 }
 
 func (c *client) GetDatabases() ([]*Database, error) {
-	req, err := c.newRequest(http.MethodGet, "/api/database", nil)
+	req, err := c.newRequest(http.MethodGet, "/api/database?include=tables", nil)
```
bsushmith, line 134: why do we need tables here?

plugins/providers/metabase/client.go
```diff
@@ -170,6 +172,134 @@ func (c *client) GetCollections() ([]*Collection, error) {
...
+	addResourceToGroup(collectionResourceGroups, groupMap, "collection")
+
+	return groups, databaseResourceGroups, collectionResourceGroups, err
+}
+
+func (c *client) fetchGroups(wg *sync.WaitGroup, groups *[]*Group, err error) {
+	defer wg.Done()
+	req, err := c.newRequest(http.MethodGet, "/api/permissions/group", nil)
+	if err != nil {
+		return
```
bsushmith, line 207: add a log here.

plugins/providers/metabase/client.go
```diff
@@ -170,6 +172,134 @@ func (c *client) GetCollections() ([]*Collection, error) {
...
+	}
+
+	_, err = c.do(req, &groups)
+	if err != nil {
+		return
```
bsushmith, line 212: add a log here.

plugins/providers/metabase/client.go
```diff
@@ -170,6 +172,134 @@ func (c *client) GetCollections() ([]*Collection, error) {
...
+	}
+}
+
+func (c *client) fetchDatabasePermissions(wg *sync.WaitGroup, resourceGroups map[string][]map[string]interface{}, err error) {
+	defer wg.Done()
+
+	req, err := c.newRequest(http.MethodGet, "/api/permissions/graph", nil)
+	if err != nil {
+		return
```
bsushmith, line 221: log here too. and everywhere where were are returning inside from a goroutine in error scenario.

plugins/providers/metabase/client.go
```diff
@@ -170,6 +172,134 @@ func (c *client) GetCollections() ([]*Collection, error) {
...
+	}
+
+	graphs := make(map[string]interface{}, 0)
+	_, err = c.do(req, &graphs)
+	if err != nil {
+		return
+	}
+
+	for groupId, r := range graphs["groups"].(map[string]interface{}) {
```
bsushmith, line 230: this block is difficult to read thru. Can we have concrete types and methods here

plugins/providers/metabase/client.go
```diff
@@ -170,6 +172,134 @@ func (c *client) GetCollections() ([]*Collection, error) {
...
+			}
+			addGroupToResource(resourceGroups, fmt.Sprintf("database:%s", dbId), groupId, err)
+		}
+	}
+}
+
+func (c *client) fetchCollectionPermissions(wg *sync.WaitGroup, resourceGroups map[string][]map[string]interface{}, err error) {
+	defer wg.Done()
+
+	req, err := c.newRequest(http.MethodGet, "/api/collection/graph", nil)
```
bsushmith, line 253: Can you make a list of constants at the top of this file for all these endpoints

plugins/providers/metabase/provider.go
```diff
@@ -45,14 +45,14 @@ func (p *provider) GetResources(pc *domain.ProviderConfig) ([]*domain.Resource,
...
+	var resourceTypes = make(map[string]bool, 0)
 	for _, rc := range pc.Resources {
-		resourceTypes = append(resourceTypes, rc.Type)
+		resourceTypes[rc.Type] = true
 	}
 
 	resources := []*domain.Resource{}
 
-	if containsString(resourceTypes, ResourceTypeDatabase) {
+	if _, ok := resourceTypes[ResourceTypeDatabase]; ok && resourceTypes[ResourceTypeDatabase] {
```
bsushmith, line 55: wouldn't this value `resourceTypes[ResourceTypeDatabase]` be true always if found in map at this line?

plugins/providers/metabase/provider.go
```diff
@@ -62,10 +62,21 @@ func (p *provider) GetResources(pc *domain.ProviderConfig) ([]*domain.Resource,
 			db.ProviderType = pc.Type
 			db.ProviderURN = pc.URN
 			resources = append(resources, db)
+
+			if _, ok := resourceTypes[ResourceTypeTable]; ok && resourceTypes[ResourceTypeTable] {
```
bsushmith, line 66: same here also.

plugins/providers/metabase/provider.go
```diff
@@ -62,10 +62,21 @@ func (p *provider) GetResources(pc *domain.ProviderConfig) ([]*domain.Resource,
...
+				tables := d.Tables
```
bsushmith, line 67: you can remove this one-time variable.

plugins/providers/metabase/provider.go
```diff
@@ -62,10 +62,21 @@ func (p *provider) GetResources(pc *domain.ProviderConfig) ([]*domain.Resource,
...
+					table.ProviderType = pc.Type
+					table.ProviderURN = pc.URN
+					resources = append(resources, table)
+				}
+			}
 		}
 	}
 
-	if containsString(resourceTypes, ResourceTypeCollection) {
+	if _, ok := resourceTypes[ResourceTypeCollection]; ok && resourceTypes[ResourceTypeCollection] {
```
bsushmith, line 79: same here also.

plugins/providers/metabase/client.go
singhvikash11: this one filter will fetch tables in the same fetch database API call

plugins/providers/metabase/client.go
bsushmith: just for consistency sake I guess. the other similar ones here use the map to share data... while this one uses slice and then gets converted map ultimately in the caller function anyway.

plugins/providers/metabase/client.go
singhvikash11: no, these are return values

plugins/providers/metabase/client.go
singhvikash11: here we are not using map because we want to selected field from /ap/permissions/group API

plugins/providers/metabase/resource.go
singhvikash11: done

plugins/providers/metabase/resource.go
singhvikash11: done

plugins/providers/metabase/client.go
singhvikash11: done

plugins/providers/metabase/provider.go
singhvikash11: I am checking this to confirm the value in the map is true.

plugins/providers/metabase/provider.go
singhvikash11: I am checking this to confirm the value in the map is true.

plugins/providers/metabase/provider.go
singhvikash11: I am checking the value should be true in the map

plugins/providers/metabase/client.go
singhvikash11: not possible, since there is no standard response from API

plugins/providers/metabase/provider.go
bsushmith: In what cases, would the key exist and the value be `false` ?

bsushmith (approved):

rahmatrhd:
@singhvikash11 I'm not seeing the new resource types (table and group) are accessible for users through grant/revoke action (CMIIW). How's the plan to make it available to the users for requesting access?

rahmatrhd (changes requested):
https://github.com/odpf/guardian/pull/162/files#diff-73adec9ef2ecd97f284a7e687b64f58d569acee4585db1711b53e2095899d5e8R171-R176
need to add rules validation for the new resource types in here as well

plugins/providers/metabase/client.go
```diff
@@ -166,11 +199,145 @@ func (c *client) GetCollections() ([]*Collection, error) {
 	if _, err := c.do(req, &collection); err != nil {
 		return nil, err
 	}
-
+	c.logger.Info(fmt.Sprintf("Fetch total %d collections from request: %v", len(collection), req.URL))
```
mabdh, line 202: With `/salt/log`, we can utilise its [kv pair logging](https://github.com/odpf/salt/blob/main/log/logger.go#L13), like this

```suggestion
	c.logger.Info(fmt.Sprintf("Fetch collections from request, "total", len(collection), "url", req.URL))
```

plugins/providers/metabase/config.go
```diff
@@ -170,6 +173,10 @@ func (c *Config) validatePermission(resourceType string, value interface{}) (*Pe
 		nameValidation = "oneof=schemas:all native:write"
 	} else if resourceType == ResourceTypeCollection {
 		nameValidation = "oneof=read write"
+	} else if resourceType == ResourceTypeTable {
+		nameValidation = "oneof=all"
+	} else if resourceType == ResourceTypeGroup {
+		nameValidation = "len=0"
```
rahmatrhd, line 179: when creating an appeal, we are asked for `email`, `resource_id`, and `role`. This makes we're allowed to enter empty `role` (`""`) which I think could be confusing. What do you think if we add one role like `member` or other else (open for discussion) for metabase group? so when creating an appeal the `role` value won't be empty

docs/docs/providers/metabase.md
```diff
@@ -85,7 +85,9 @@ resources:
 ### `MetabaseResourceType`
 
 - `database`
+- `table`
 - `collection`
+- `group`
 
 ### `MetabaseResourcePermission`
```
rahmatrhd, line 92: please add possible values for permissions for table and group

docs/docs/providers/metabase.md
```diff
@@ -85,7 +85,9 @@ resources:
 ### `MetabaseResourceType`
```
rahmatrhd, line 86: please add resource config example for table and group in the `#### Example`

docs/docs/providers/metabase.md
```diff
@@ -72,6 +72,22 @@ resources:
...
+        name: Viewer
+        permissions:
+          - all
+  - type: group
+    policy:
+      id: policy_id
+      version: 1
+    roles:
+      - id: member
+        name: Member
```
rahmatrhd, line 90:
```yaml
permission: 
  - member
```
permission `member` should be required, right? because we have just added the validation in the `validatePermission()`

singhvikash11: yes, done

rahmatrhd (approved):
LGTM

PR #164, opened 2022-05-11
feat(cmd): add sub-command for running jobs

### LINKED ISSUE #140: feat: run guardian jobs from cli
Right now we're still running the jobs from the server, ideally we should expose each job, and let the user decide how the job is triggered. In that case, we can create a CLI command to expose the jobs such as:

$ guardian jobs {{job_name}} run

rahmatrhd:
closes #140 

usage: `guardian jobs run <job_name>`
example: `guardian jobs run fetch_resources`
valid job names: `fetch_resources`, `appeal_expiration_reminder`, `appeal_expiration_revocation`

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -9,6 +9,12 @@ import (
 	"github.com/odpf/salt/config"
 )
 
+type Jobs struct {
+	FetchResourcesInterval             string `mapstructure:"fetch_resources_interval" default:"0 */2 * * *"`
+	RevokeExpiredAccessInterval        string `mapstructure:"revoke_expired_access_interval" default:"*/20 * * * *"`
+	ExpiringAccessNotificationInterval string `mapstructure:"expiring_access_notification_interval" default:"0 9 * * *"`
+}
+
 type Config struct {
 	Port                       int              `mapstructure:"port" default:"8080"`
 	EncryptionSecretKeyKey     string           `mapstructure:"encryption_secret_key"`
--- a/cmd/job.go
+++ b/cmd/job.go
@@ -42,7 +42,7 @@ func runJobCmd() *cobra.Command {
 			$ guardian job run appeal_expiration_reminder
 			$ guardian job run appeal_expiration_revocation
 		`),
-		Args: cobra.OnlyValidArgs,
+		Args: cobra.ExactValidArgs(1),
 		ValidArgs: []string{
 			"fetch_resources",
 			"appeal_expiration_reminder",
--- a/cmd/job.go
+++ b/cmd/job.go
@@ -13,12 +13,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func JobsCmd() *cobra.Command {
+func JobCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "jobs",
-		Short: "Manage jobs",
+		Use:     "job",
+		Aliases: []string{"jobs"},
+		Short:   "Manage jobs",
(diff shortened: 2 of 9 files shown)


cmd/jobs.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/odpf/guardian/internal/crypto"
+	"github.com/odpf/guardian/jobs"
+	"github.com/odpf/guardian/plugins/notifiers"
+	"github.com/odpf/salt/log"
+	"github.com/spf13/cobra"
+)
+
+func JobsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "jobs",
```
ravisuhag, line 18: @rahmatrhd to make it consistent with rest of the commands, it should be `job` and we can add `jobs` as an alias.

mabdh:
@rahmatrhd for `config.yaml` file, should we put it in the root project under the name `config.yaml.example` ?

mabdh:
@rahmatrhd for `config.yaml` file, should we put it in the root project under the name `config.yaml.example` ?

cmd/job.go
```diff
@@ -0,0 +1,102 @@
...
+				notifier,
+			)
+
+			jobs := map[string]func() error{
+				"fetch_resources":              handler.FetchResources,
+				"appeal_expiration_reminder":   handler.AppealExpirationReminder,
+				"appeal_expiration_revocation": handler.RevokeExpiredAppeals,
+			}
+
+			job := jobs[args[0]]
```
mabdh, line 93: This throws panic if we run `guardian job run` only.

```
panic: runtime error: index out of range [0] with length 0
```

mabdh (changes requested):

mabdh:
When there is a typo with the job name it shows like this
```bash
$ guardian job run fetch_resources1111          
invalid argument "fetch_resources1111" for "guardian job run"
```

should we add more information e.g. suggestions/shows some supported jobs?

app/config.go
```diff
@@ -9,6 +9,12 @@ import (
 	"github.com/odpf/salt/config"
 )
```
mabdh, line 10: @rahmatrhd for config.yaml file, should we put it in the root project under the name config.yaml.example ?

cmd/job.go
```diff
@@ -0,0 +1,102 @@
...
+
+	return cmd
+}
+
+func runJobCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "run",
+		Short: "Fire a specific job",
+		Example: heredoc.Doc(`
+			$ guardian job run fetch_resources
```
mabdh, line 41: When there is a typo with the job name it shows like this
```
$ guardian job run fetch_resources1111          
invalid argument "fetch_resources1111" for "guardian job run"
```
should we add more information e.g. suggestions/shows some supported jobs?

rahmatrhd: couldn't find a way how to add custom message for arg validation in cobra. @mabdh do you have any suggestions?

app/config.go
rahmatrhd: don't remember putting config.yaml here, I think it is used for documentation, and also because it is a config for "app", another config we have is for "cmd", maybe to segregate those CMIIW

cmd/job.go
mabdh: yes there is no simple solution with cobra for this, I think we can just keep it like this since this not really crucial to do.

mabdh (approved):

PR #167, opened 2022-05-19
feat: initial implementation of metabase group migration into guardian

singhvikash11:


cmd/migration.go
```diff
@@ -1,131 +1,202 @@
...
+)
 
-func metabaseMigrationCmd() *cobra.Command {
-	return &cobra.Command{
-		Use:   "metabase",
-		Short: "Metabase migration",
+func MigrationCmd(config *Config) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "migrate",
+		Short: "Guardian migration",
```
ravisuhag, line 26: @rahmatrhd @singhvikash11 I think we need to think of a better approach than attaching it on migrate. One possible suggestion could be to call it import and scope it with the provider itself.

singhvikash11: @ravisuhag this is a good idea, we need to implement migrate for bigquery provider and other providers too. Then we could expose it as a sub-command on the provider

singhvikash11: @ravisuhag @rahmatrhd If you folks are okay with Metabase migrate for now then I can move this as sub-command to provider.

ravisuhag: @singhvikash11 What do you mean okay with metabase provider?

Also, i would recommend checking out https://github.com/GoogleCloudPlatform/terraformer for some inspiration on how command could look like.

singhvikash11: @ravisuhag we already have the command `guardian provider create` command to register the resources with guardian. Let's use the following command to import ACL of existing resources from a source system:

`guardian provider import <provide-name> acl `
`guardian provider import <provide-name> acl -r <resources>`

what do you guys think? cc @rahmatrhd

ravisuhag: @singhvikash11 What does `acl` represent in this command?

singhvikash11: @ravisuhag Bigquery, Metabase resources like table, dataset, collection etc defined by a set of permissions associated with a set of people, group, service-account. By importing access-control of these resources into our system we'll replicate it in the model of appeal-resource in Guardian.

ravisuhag: @singhvikash11 Yes, understood. But what i meant is that we don't need acl word in the comamnd right? the command can plainly be `guardian provider import <provide-name> -r <resources>` where `-r` flag will take the kind of resources you want to import from that provider.

singhvikash11: @ravisuhag This makes sense, we can go ahead with this command. 

cc @rahmatrhd

ravisuhag:
@singhvikash11 @rahmatrhd After provider import is done. Can we move this to use that contract and update the PR?

ravisuhag:
@rahmatrhd @bsushmith Is there any plan to pick this up?

PR #176, opened 2022-06-10
fix(jobs): use createdBy instead of accountID for appeal expiry reminder

### LINKED ISSUE #175: Expiry reminders do not work for `serviceAccount` appeals.
**Describe the bug**
Expiry reminders do not work for appeals created for a `serviceAccount`. 

This [snippet](https://github.com/odpf/guardian/blob/main/jobs/appeal_expiration_reminder.go#L40) in AppealExpirationReminder method - 

```golang

		for _, a := range appeals {
			notifications = append(notifications, domain.Notification{
				User: a.AccountID,
				Message: domain.NotificationMessage{
					Type: domain.NotificationTypeExpirationReminder,

```

should ideally be using `appeal.CreatedBy` instead of AccountID for user, as accountID would have serviceAccount and it would not have any slackID tied to it to send a notification to. 

This issue doesn't exist when sending notification for revoke, since revoke flow [uses](https://github.com/odpf/guardian/blob/main/core/appeal/service.go#L475) appeal.AccountID for notifications.

bsushmith:
fix #175

### DIFF
--- a/jobs/appeal_expiration_reminder.go
+++ b/jobs/appeal_expiration_reminder.go
@@ -37,7 +37,7 @@ func (h *handler) AppealExpirationReminder(ctx context.Context) error {
 		var notifications []domain.Notification
 		for _, a := range appeals {
 			notifications = append(notifications, domain.Notification{
-				User: a.AccountID,
+				User: a.CreatedBy,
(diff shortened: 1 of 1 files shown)


rahmatrhd (approved):
LGTM

PR #186, opened 2022-06-23
fix(appeal): additional appeal policy override

rahmatrhd:
### bug:
policy example:
```yaml
...
requirements:
  on:
    resource_id: 123
  appeals:
    - resource:
         id: 456
       role: viewer
       policy:
         id: policy-x # overriding policy
         version: 99
```
the policy (policy-x@99) specified above won't get applied for the additional appeal creation for resource_id:456

this PR is for fixing that bug

### DIFF
--- a/core/appeal/service.go
+++ b/core/appeal/service.go
@@ -809,9 +809,11 @@ func (s *Service) CreateAccess(ctx context.Context, a *domain.Appeal) error {
 		policy = p
 	}
 
-	// TODO: don't handle appeal requirements if current appeal already an additional appeal
-	if err := s.handleAppealRequirements(ctx, a, policy); err != nil {
-		return fmt.Errorf("handling appeal requirements: %w", err)
+	isAdditionalAppealCreation, _ := ctx.Value(ContextKeyIsAdditionalAppealCreation{}).(bool)
+	if !isAdditionalAppealCreation {
+		if err := s.handleAppealRequirements(ctx, a, policy); err != nil {
+			return fmt.Errorf("handling appeal requirements: %w", err)
+		}
 	}
 
 	if err := s.providerService.GrantAccess(ctx, a); err != nil {
--- a/core/appeal/service.go
+++ b/core/appeal/service.go
@@ -36,6 +36,8 @@ const (
 
 var TimeNow = time.Now
 
+type ContextKeyIsAdditionalAppealCreation struct{}
+
 type repository interface {
 	BulkUpsert([]*domain.Appeal) error
 	Find(*domain.ListAppealsFilter) ([]*domain.Appeal, error)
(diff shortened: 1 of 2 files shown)


core/appeal/service_test.go
```diff
@@ -768,6 +769,84 @@ func (s *ServiceTestSuite) TestCreate() {
...
+			s.mockIAMClient.On("GetUser", input.AccountID).Return(map[string]interface{}{}, nil)
+			s.mockApprovalService.On("AdvanceApproval", mock.Anything, mock.Anything).Return(nil)
+			s.mockRepository.On("BulkUpsert", mock.Anything).Return(nil).Once()
+			s.mockNotifier.On("Notify", mock.Anything).Return(nil).Once()
+			s.mockAuditLogger.On("Log", mock.Anything, appeal.AuditKeyBulkInsert, mock.Anything).Return(nil).Once()
+
+			err := s.service.Create(ctx, []*domain.Appeal{input})
+
+			s.NoError(err)
+			s.Equal("test-approval", input.Approvals[0].Name)
```
mabdh, line 847: We can add `.AssertExpectations(s.T())` for all mock object to make sure all mocks are executed.

core/appeal/service.go
```diff
@@ -776,6 +786,7 @@ func (s *Service) handleAppealRequirements(ctx context.Context, a *domain.Appeal
 					additionalAppeal.PolicyID = aa.Policy.ID
 					additionalAppeal.PolicyVersion = uint(aa.Policy.Version)
 				}
+				ctx = context.WithValue(ctx, ContextKeyIsAdditionalAppealCreation{}, true)
```
mabdh, line 789: can we pass the flag via agument instead? 
i don't think passing flag by context is better when we could just pass it via argument.

rahmatrhd: we can but the reason I'm using context is to keep the `appealService.Create` method clean. I don't think it should have the flag added in the param CMIIW

mabdh: @rahmatrhd passing the required value of a function via context would make the function less readable. Now the `Create` function does have 2 possible behaviour to populate policy. Adding the flag in the function argument would make it clearer what to expect.

Since the context.Value is an interface. We also will lose the `compile-type` checking benefit. 

Other alternative would be create a similar function maybe named as `Append` but this might lead to some duplicated code.

The other alternative, we could add a new variadic options in the argument and if needed user can add the flag.

I think 'clean method' does not really give benefit here if it is less readable.

bsushmith: variadic options feels like a cleaner approach in this situation. Is clean and readable.

rahmatrhd: @mabdh apart from this, can you approve this PR?

bsushmith (approved):

ravisuhag (approved):

Issue #191, opened 2022-06-24
Ability to enable / disable jobs running in server through config


bsushmith:
**Summary**
Guardian supports running jobs as commands. That means one should be able to run the jobs separately in a cron. I should be able to disable running jobs in server, so that I can run them separately outside the server in a cron.

This also resolves one other issue - Whenever guardian jobs run in the server, if we have multiple k8s pods - a user will receive multiple notifications at the same time from each pod. Running the jobs separately would ensure clean user experience and also separate out the cron jobs.

**Proposed solution**
Add enable/disable flag like below  in server config
```yaml
JOBS:
  FETCH_RESOURCES:
    Enable: True
    Interval: "0 */2 * * *"
  REVOKE_EXPIRED_ACCESS:
    Enable: True
    Interval: "*/20 * * * *"
  EXPIRING_ACCESS_NOTIFICATION:
    Enable: True
    Interval: "0 9 * * *"

```

**Additional context**
Guardian version >= `v0.2.9`

singhvikash11:
@bsushmith can k8s deployments pass different configs to pods.

ravisuhag:
@bsushmith @Chief-Rishab Can we please update the docs for this change?

PR #193, opened 2022-06-25
feat(jobs)!: add config to enable/disable job

bsushmith:
Fixes: #191 

BREAKING CHANGE: jobs can now be disabled/enabled. If enabled, the job specific schedule has to be passed to the config as the default interval is same for all jobs.


singhvikash11 (approved):

internal/server/config.go
```diff
@@ -9,10 +9,24 @@ import (
...
+const (
+	FetchResources             JobType = "fetch_resources"
+	ExpiringAccessNotification JobType = "appeal_expiration_reminder"
+	RevokeExpiredAccess        JobType = "appeal_expiration_revocation"
+)
+
+type JobConfig struct {
+	JobType  JobType
+	Enabled  bool   `mapstructure:"enabled" default:"true"`
+	Interval string `mapstructure:"interval" default:"0 9 * * *"`
```
rahmatrhd, line 23: instead of setting `0 9 * * *` as the default value for every `JobConfig` can we keep the default values for each job interval manually?

rahmatrhd (changes requested):

internal/server/config.go
bsushmith: yeah. I wanted to do that to keep backward compatibility, but it meant  - I would have to model these as anonymous structs, and not as a single JobConfig struct if I need that default interval in tags. 

We can add that custom default intervals in the function - `func fetchJobsToRun(config *Config) []JobConfig { ... }`

ravisuhag: @rahmatrhd @bsushmith what is our final decision on this?

rahmatrhd: @bsushmith what do you think if we add a few lines of code for assigning the default values for these three jobs?

rahmatrhd (approved):

PR #197, opened 2022-06-28
fix(appeal): additional appeals fix

rahmatrhd:


mabdh:
@rahmatrhd is it duplicated with https://github.com/odpf/guardian/pull/186 ?

PR #200, opened 2022-07-04
feat: access revocation for dormant account

singhvikash11:
### Revoke appeal access for a dormant account.

Guardian appeal-approval workflow takes identity providers(IAM) like Shied, Ldap, and HTTP to fetch user details like name, title, manager, department etc.

As a solution we are proposing the following steps:

- identity providers have user status as metadata like accountStatus=active

- a scheduled job in Guardian
   > to fetch all users having active appeals
   > fetch the status of a user using IAM
   > revoke appeal access if a user is not active

It'll fix  [issue](https://github.com/odpf/guardian/issues/202)


ravisuhag:
@singhvikash11 can you corresponding issue for it and document your approach for it first.

domain/appeal.go
```diff
@@ -120,6 +120,7 @@ type ApprovalAction struct {
 type ListAppealsFilter struct {
 	CreatedBy                 string    `mapstructure:"created_by" validate:"omitempty,required"`
 	AccountID                 string    `mapstructure:"account_id" validate:"omitempty,required"`
+	AccountType               string    `mapstructure:"account_type" validate:"omitempty,required"`
```
ravisuhag, line 123: @singhvikash11 Does account_type reflect that it is active or inactive? If yes, should it be called `account_status`? And can we also think about what possible values it can have?

domain/iam.go
```diff
@@ -21,4 +21,5 @@ type IAMManager interface {
 // IAMClient interface
 type IAMClient interface {
 	GetUser(id string) (interface{}, error)
+	IsActiveUser(id string) (bool, error)
```
ravisuhag, line 24: @singhvikash11 If I understand this correctly, this function will be used to check if the user is active or not on demand? Does `GetUser` return this status? Any reason for us to have this as a separate method?

internal/server/config.go
```diff
@@ -10,9 +10,10 @@ import (
 )
 
 type Jobs struct {
-	FetchResourcesInterval             string `mapstructure:"fetch_resources_interval" default:"0 */2 * * *"`
-	RevokeExpiredAccessInterval        string `mapstructure:"revoke_expired_access_interval" default:"*/20 * * * *"`
-	ExpiringAccessNotificationInterval string `mapstructure:"expiring_access_notification_interval" default:"0 9 * * *"`
+	FetchResourcesInterval                    string `mapstructure:"fetch_resources_interval" default:"0 */2 * * *"`
+	RevokeExpiredAccessInterval               string `mapstructure:"revoke_expired_access_interval" default:"*/20 * * * *"`
+	ExpiringAccessNotificationInterval        string `mapstructure:"expiring_access_notification_interval" default:"0 9 * * *"`
+	RevokeDormantAccountAppealExpiredInterval string `mapstructure:"expiring_access_notification_interval" default:"0 0 * * *"`
```
ravisuhag, line 16: @singhvikash11 Dormant access can be defined in what ways I guess, so we can keep this specific by calling it `RevokeInActiveUserAccess`

plugins/identities/http.go
```diff
@@ -148,6 +148,20 @@ func (c *HTTPClient) GetUser(userID string) (interface{}, error) {
...
 }
 
+func (c *HTTPClient) IsActiveUser(userID string) (bool, error) {
+	userDetails, err := c.GetUser(userID)
+	if err != nil {
+		return false, err
+	}
+
+	if userDetailsMap, ok := userDetails.(map[string]interface{}); ok {
+		if accountStatus, ok := userDetailsMap[accountStatus]; ok {
```
ravisuhag, line 158: @singhvikash11 This `accountStatus ` key is hardcoded. We can not assume that IAM will always have this as key. They can actually have a different key to denote this. Also Guardian should decouple this from IAM I think. Guardian can manage its status for a user if needed and use that to revoke access. The status in Guardian can then be used to refresh from IAM and revoke access if needed. WDYT?

singhvikash11: slightly disagreement here and I am visualising this as part of the first implementation:
`iam:
  config:
   provider: http
  schema:
      account_status: status == 'InActive'`

domain/iam.go
singhvikash11: Having a separate `IsActiveUser ` function will allow users to write some complex custom functions to connect external systems etc. If you say we can move into `GetUser`

domain/appeal.go
rahmatrhd: `account_type` was already there in appeal model, it is for specifying the type of `account_id` e.g.: `user` or `serviceAccount`

plugins/identities/http.go
rahmatrhd: instead, we can add new config for determining account status field, e.g.:
```yaml
iam:
  provider: http
  config:
    url: ...
    schema:
      name: fullName
      foo: bar
      ...
    account_status_key: is_active # boolean expected
    # OR
    account_status: status == 'active'
```

with user object example as following
```json
{
  "fullName": "John Doe",
  "status": "active",
  "is_active": true,
  ...
}
```

singhvikash11: I like this approach, should I go ahead and implement this

@rahmatrhd @ravisuhag wdyt?

ravisuhag: @singhvikash11 can you update the changes discussed so far on the issue to cross-check the final approach?

singhvikash11: @rahmatrhd @ravisuhag As we discussed, the approach should be:

- A cron job to trigger this flow.
- Job will fetch all active appeals and group them based on the appeal's user, policy_id and policy_version.
- For each group, it'll evaluate user account status as either `true` or `false` based on the status expression configure in the policy.
- For each user with account status `false`, the job will revoke all appeals for that user. 

The job will show o/p like:

> INFO[0000] running inactive user appeal revoke job      
INFO[0001] successful appeal revocation                  count=0 ids="[]" user=vikash.singh@gojek.com
INFO[0001] failed appeal revocation                      count=0 ids="[]" user=vikash.singh@gojek.com

> INFO[0000] running inactive user appeal revoke job      
INFO[0003] appeal revoked                                id=e094bbec-e826-469f-9685-c780c5de76d8
INFO[0003] successful appeal revocation                  count=1 ids="[e094bbec-e826-469f-9685-c780c5de76d8]" user=vikash.singh@gojek.com
INFO[0003] failed appeal revocation                      count=0 ids="[]" user=vikash.singh@gojek.com

singhvikash11: Sample policy:

>
{
    "id": "sample_policy",
    "steps": [],
    "iam": {
        "provider": "http",
        "config": {
            },
        "schema": {
            "job_title": "title",
            "manager_email": "managerEmail"
        },
        **"account_status":"$user.accountStatus == 'active'"**
    }
}

ravisuhag:
@rahmatrhd @singhvikash11 https://github.com/odpf/guardian/issues/202#issuecomment-1184727135

PR #201, opened 2022-07-05
feat(provider): introduce no-op provider

### LINKED ISSUE #178: Support noop provider
**Summary**
To create an appeal guardian requires a provider to be registered. But in some cases, Guardian users might just want to take advantage of policy workflow and not load resources in Guardian. They might want to just use Guardian APIs for approval workflows and appeal management. This can also allow users to locally test Guardian easily without configuring an actual provider. 

**Proposed solution**
Guardian can support a no-op provider which will require be placeholder provider with no resources. Users would be able to configure a policy on no-op providers and raise an appeal. In such cases, users would typically integrate with Guardian APIs and might want to manage resources for which appeals are being created on the application side itself.

rahmatrhd:
closes #178

### DIFF
--- a/domain/provider.go
+++ b/domain/provider.go
@@ -15,6 +15,8 @@ const (
 	ProviderTypeTableau = "tableau"
 	// ProviderTypeGCloudIAM is the type name for Google Cloud IAM provider
 	ProviderTypeGCloudIAM = "gcloud_iam"
+	// ProviderTypeNoOp is the type name for No-Op provider
+	ProviderTypeNoOp = "no_op"
 )
 
 // Role is the configuration to define a role and mapping the permissions in the provider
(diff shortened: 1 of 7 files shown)


domain/provider.go
```diff
@@ -15,6 +15,8 @@ const (
 	ProviderTypeTableau = "tableau"
 	// ProviderTypeGCloudIAM is the type name for Google Cloud IAM provider
 	ProviderTypeGCloudIAM = "gcloud_iam"
+	// ProviderTypeNoOp is the type name for No-Op provider
+	ProviderTypeNoOp = "no_op"
```
ravisuhag, line 19: @rahmatrhd Let's keep it plain(without a dash) `noop`

ravisuhag (approved):

PR #220, opened 2022-07-18
fix: keep permissions state in appeal

### LINKED ISSUE #205: bug: appeal grant/revoke after the role is changed/removed from the provider config
### **Description**
User (admin) can edit/remove roles from a provider at any time, and the previously created access with that role will lost the information of what are the permissions were in that role.

#### Example 1: Approving appeal
1. Create a provider with a role `viewer` in the `roles`
2. Create an appeal to one of the resources with `role=viewer`
3. Edit the provider and remove that `viewer` role
4. Approve the appeal

#### Example 2: Revoking appeal
1. Create a provider with a role `viewer` in the `roles`
2. Create an appeal to one of the resources with `role=viewer`
3. Approve the appeal
4. Edit the provider and remove that `viewer` role
5. Revoke the appeal

### **Expected behavior**
Granting/revoking access after the creation should be working regardless the changes in the provider (especially edit/remove changes in the roles)

### **Additional context**
- bigquery roles such as `roles/bigquery.dataEditor` and `roles/bigquery.metadataEditor` are called `permissions` in guardian
- In guardian, user (admin) can define their own custom roles from the provider config. Let’s say one role named `bigquery-admin` and it has multiple permissions (bigquery roles) such as `roles/bigquery.dataEditor` and `roles/bigquery.metadataEditor`. So once user requested an appeal with `bigquery-admin`, they will get both permissions granted to the resource in the bigquery

### **Proposed solution**
Need to store the list of permissions in the appeal object (probably a new field named "permissions") on the appeal creation. This `permissions` field will be used in the future for granting/revoking the access to the provider, regardless of the changes in the provider.

rahmatrhd:
closes #205 

proton PR: https://github.com/odpf/proton/pull/162


internal/store/postgres/migrations/000001_create_providers_table.down.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE IF EXISTS "providers";
```
mabdh, line 1: do we just adding a migration scripts now?
if yes I think this migration could be on separate PR so it will be in the separate commit in the main branch

internal/store/postgres/migrations/000005_add_permissions_to_appeal.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "appeals" DROP COLUMN IF EXISTS "permissions";
```
mabdh, line 1: If we just added a migration script, we don't need this alter table command.

internal/store/postgres/migrations/000001_create_providers_table.up.sql
```diff
@@ -0,0 +1,15 @@
...
+  "urn" text,
+  "config" JSONB,
+  "created_at" timestamptz,
+  "updated_at" timestamptz,
+  "deleted_at" timestamptz,
+  PRIMARY KEY ("id")
+);
+
+CREATE INDEX IF NOT EXISTS "idx_providers_deleted_at" ON "providers" ("deleted_at");
+CREATE UNIQUE INDEX IF NOT EXISTS "provider_index" ON "providers" ("type", "urn");
```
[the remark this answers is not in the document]
rahmatrhd: not sure if we can do this, since these migration scripts are created to replace gorm's auto migration, so the table schemas, indexes, etc. are already present in the db.

internal/store/postgres/migrations/000005_add_permissions_to_appeal.down.sql
rahmatrhd: since this is for replacing auto migration, migration scripts version 1-4 are compatible sql scripts with the existing schema in the db, this version 5 is the addition to add a column

core/provider/common.go
```diff
@@ -11,3 +11,19 @@ func GetRoles(pc *domain.ProviderConfig, resourceType string) ([]*domain.Role, e
 
 	return nil, ErrInvalidResourceType
 }
+
+type PermissionManager struct{}
```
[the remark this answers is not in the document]
mabdh: Using `PermissionManager` interface would give benefit if there are possible ways to do `GetPermission`. One of the way is the `type PermissionManager struct`.

If the logic of `GetPermissions` is the same for all providers, that is a strong sign that this does not to be interfaced, hence not needed a specific struct. This will make code simpler and reduce the dependency for all providers `e.g. BigQueryProvider does not have PermissionManager inside it`. We can just call `package.Function` (provider.GetPermission). If we do this, we don't need to mock  `GetPermission` in the test.

internal/store/postgres/migrations/000001_create_providers_table.down.sql
mabdh: To decouple DB migration changes from app logic changes. Commit better to be as granular as possible and I see DB migration and app logic as two different changes. Both needs to be tested differently. If we need to revert, it is easier to do it if DB migration has its own commit.

internal/store/postgres/migrations/000005_add_permissions_to_appeal.down.sql
mabdh: Yes I know but Is the current migration version tracked in the DB with auto migrate?. If not we could merge the column changes into the table creation migration itself. To avoid complex migration script.

rahmatrhd: It's actually the other way around, gorm automigrate doesn't track the schema version because it only uses the latest model and generates sql if there are any changes to the model. And with moving to go-migrate, we need to create migration scripts that match the existing schema in db first (in this case .sql version 1-4) so that when it's deployed and runs the migration script with go-migrate the first time, it won't have conflict and adds `schema_migrations` table as well.

Also, merging the column changes into the table creation won't be possible because the existing db in production already has "appeals" table, and "permissions" column needs to be added and can be done only by altering the table itself. 

The reason we're moving to go-migrate it's because we need to backfill the old data to have a proper value for the newly added "permissions" column which is not supported by gorm auto migrate

internal/store/postgres/migrations/000001_create_providers_table.down.sql
rahmatrhd: we can decouple gorm to go-migrate migration replacement into separate PR, but even reverting that commit in the future would cause an error after this "permissions" column changes applied because this changes depends on go-migrate

internal/store/postgres/migrations/000005_add_permissions_to_appeal.down.sql
mabdh: @rahmatrhd Yes exactly that is the point. I think to make it clear (this is related with https://github.com/odpf/guardian/pull/220#discussion_r927369892).

We can have a different migration scripts PR (adding a `permission` column is not included) and merge it first. This PR could have an additional of migration script to alter table add column `permission`.

internal/store/postgres/migrations/000001_create_providers_table.down.sql
mabdh: As discussed in this [comment](https://github.com/odpf/guardian/pull/220#discussion_r927373822). This PR could just have a migration file `000005_add_permissions_to_appeal.down.sql`. The version 1-4 could be in a separate PR and being merge first.

core/provider/common.go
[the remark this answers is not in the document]
rahmatrhd: yes, the only thing right now is gcloud_iam provider has different implementation on how to retrieve permissions: https://github.com/odpf/guardian/blob/069f69b9a912fe6872b7e0773a23b5345349db73/plugins/providers/gcloudiam/provider.go#L141-L144

so the possible options are to use `type PermissionManager struct` like the current one, or have `provider.GetPermissions()` being called in every provider plugins as you suggested

rahmatrhd:
@mabdh cherry-picked the migrator changes into this PR https://github.com/odpf/guardian/pull/228

mabdh:
@rahmatrhd I have tested with `no-op` provider, since `no-op` cannot have `permission` in its role, creating appeal with `no-op` provider is failing because constraint in DB column `permission` cannot be `NULL`.

```
failed to create appeal: inserting appeals into db: ERROR: null value in column \"permissions\" violates not-null constraint (SQLSTATE 23502)
```

rahmatrhd:
@mabdh I've fixed it by returning empty array if provider `permissions` is nil

mabdh:
@rahmatrhd that is still being translated to `NULL` by gorm and `permission` column is non-nullable.

rahmatrhd:
@mabdh it's fixed now

mabdh (approved):

ravisuhag (approved):

PR #222, opened 2022-07-18
feat: use variadic options for appeal creation

### LINKED ISSUE #221: use variadic options instead of context to pass around additional appeal information
**Summary**
use variadic options for create appeal instead of using context variable to pass around additional appeal flag.

**Proposed solution**
Add variadic options to `Create` appeal to infer if the appeal is an additional appeal or not.

**Additional context**
https://github.com/odpf/guardian/pull/186#discussion_r908020672

bsushmith:
Closes #221


core/appeal/service.go
```diff
@@ -78,6 +76,18 @@ type auditLogger interface {
 	Log(ctx context.Context, action string, data interface{}) error
 }
 
+type CreateAppealOption func(*createAppealOptions)
+
+type createAppealOptions struct {
+	IsAdditionalAppeal bool
+}
+
+func WithAdditionalAppealOption() CreateAppealOption {
```
mabdh, line 85:
```suggestion
func CreateWithAdditionalAppeal() CreateAppealOption {
```
better to scoped it specific to `Create` action

rahmatrhd: @mabdh can we just rely on the returned type `CreateAppealOption` instead of changing the func name?

core/appeal/service_test.go
```diff
@@ -841,7 +839,7 @@ func (s *ServiceTestSuite) TestCreate() {
 			s.mockNotifier.On("Notify", mock.Anything).Return(nil).Once()
 			s.mockAuditLogger.On("Log", mock.Anything, appeal.AuditKeyBulkInsert, mock.Anything).Return(nil).Once()
 
-			err := s.service.Create(ctx, []*domain.Appeal{input})
+			err := s.service.Create(context.TODO(), []*domain.Appeal{input}, appeal.WithAdditionalAppealOption())
```
rahmatrhd, line 842: can we use `context.Background()` instead here? since this call is the top level execution in this case

core/appeal/service.go
bsushmith: @mabdh makes sense. Made the change accordingly.

mabdh (approved):

Issue #218, opened 2022-07-12
Able to register gcp iam roles as resources


utsav-agarwal:
**Summary**
Right now, To Get the access to a Particular role, the code is been  added in terracode for it.
We trying to solve it through guardian by exposing project roles as the guardian resources and thus requesting access through console by hitting Guardian api's

**Proposed solution**
Adding Provider config in config store having resources list as the Iam roles and thus calling GCP Project roles api , filtering out the roles mentioned in provider resource list. 

@singhvikash11 @bsushmith

rahmatrhd:
the current [gcloud_iam provider](https://odpf.github.io/guardian/docs/providers/gcloud_iam) already solves this use case. Users can select a resource that is a gcp project and a role (fetched from gcp iam roles). The only missing part right now is for filtering the roles which also already raised here https://github.com/odpf/guardian/issues/184

utsav14nov:
@rahmatrhd 
We are categorising roles by business use cases which can de decided.
There can be these category types and each will have the roles assigned to it.

Following is the sample policy

```
type: gcloud_iam
urn: g-gopay-data-access
credentials:
  service_account_key: secret_key
  resource_name: projects/g-gopay-data-access
appeal:
  allow_permanent_access: true
  allow_active_access_extension_in: 24h
resources:
  - type: category-1 (Bigquery)
    policy:
      id: gofin_iam_policy
      version: 7
    roles:
      - id: roles/bigquery.admin
        name: BigQuery Admin
      - id: roles/bigquery.dataEditor
        name: BigQuery Data Editor
      - id: roles/bigquery.dataOwner
        name: BigQuery Data Owner
  - type: category-2 (Custom)
    policy:
      id: gofin_iam_policy
      version: 7
    roles:
      - id: projects/integration/roles/project.iamManager
        name: Project Iam Role Manager
  - type: category-3 (Api gateway)
    policy:
      id: gofin_iam_policy
      version: 7
    roles:
      - id: roles/apigateway.admin
        name: ApiGateway Admin
      - id: roles/apigateway.viewer
        name: ApiGateway Viewer
      - id: roles/apigateway.viewer
        name: ApiGateway Viewer
```
        
User has to select the Category first (This will be taken from type field in resource list) and then the corresponding roles list inside that category can be populated for the access.

CC: @singhvikash11 @bsushmith

rahmatrhd:
[...]

PR #223, opened 2022-07-21
feat: register Gcloud Iam roles as resources of guardian

utsav14nov:


bsushmith:
@utsav-agarwal can you fix the test cases and lint errors

rahmatrhd:
@utsav14nov need to understand the use case of this, let's discuss in https://github.com/odpf/guardian/issues/218

utsav14nov:
> 



> @utsav-agarwal can you fix the test cases and lint errors

@bsushmith Test cases and lint is fixed

docs/docs/providers/gcloud_iam.md
```diff
@@ -48,6 +64,13 @@ resources:
     policy:
       id: my_policy
       version: 1
+    roles:
```
bsushmith, line 67: Can you add a section at the bottom explaining the way permissions are configured here similar to how it's done in the `bigquery` doc

plugins/providers/gcloudiam/provider.go
```diff
@@ -50,6 +52,46 @@ func (p *Provider) GetResources(pc *domain.ProviderConfig) ([]*domain.Resource,
...
+	}
+
+	for _, rc := range pc.Resources {
+		for _, ro := range rc.Roles {
+			for _, p := range ro.Permissions {
+				permission := fmt.Sprint(p)
+				_, ok := rolesMap[permission]
+				if ok != true {
+					return nil, fmt.Errorf("%v: %v", ErrInvalidProjectRole, permission)
+				}
```
bsushmith, line 90:
```suggestion
				
				if _, ok := rolesMap[permission]; !ok {
					return nil, fmt.Errorf("%v: %v", ErrInvalidProjectRole, permission)
				}
```

plugins/providers/gcloudiam/provider.go
```diff
@@ -75,11 +117,20 @@ func (p *Provider) GrantAccess(pc *domain.ProviderConfig, a *domain.Appeal) erro
...
+						permission := fmt.Sprint(p)
+						if err := client.GrantAccess(a.AccountType, a.AccountID, permission); err != nil {
+							if !errors.Is(err, ErrPermissionAlreadyExists) {
+								return err
+							}
+						}
+					}
+					break
+				}
 			}
```
bsushmith, line 133: Can you refactor this block? this is highly nested and difficult to understand.

plugins/providers/gcloudiam/provider.go
```diff
@@ -99,11 +150,20 @@ func (p *Provider) RevokeAccess(pc *domain.ProviderConfig, a *domain.Appeal) err
...
+					for _, p := range ro.Permissions {
+						permission := fmt.Sprint(p)
+						if err := client.RevokeAccess(a.AccountType, a.AccountID, permission); err != nil {
+							if !errors.Is(err, ErrPermissionNotFound) {
+								return err
+							}
+						}
+					}
+					break
+				}
```
bsushmith, line 165: Can you refactor this block too? this is highly nested and difficult to understand.

plugins/providers/gcloudiam/provider.go
```diff
@@ -50,6 +52,46 @@ func (p *Provider) GetResources(pc *domain.ProviderConfig) ([]*domain.Resource,
...
+		roles = append(roles, &domain.Role{
+			ID:          r.Name,
+			Name:        r.Title,
+			Description: r.Description,
+		})
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
bsushmith, line 76: This is not necessary. Please remove.

plugins/providers/gcloudiam/client.go
```diff
@@ -124,13 +124,13 @@ func (c *iamClient) RevokeAccess(accountType, accountID, role string) error {
 	member := fmt.Sprintf("%s:%s", accountType, accountID)
 	for _, b := range policy.Bindings {
 		if b.Role == role {
-			var removeIndex int
+			var removeIndex int = -1
```
bsushmith, line 127:
```suggestion
			removeIndex := -1
```

docs/docs/providers/gcloud_iam.md
utsav14nov: Done

plugins/providers/gcloudiam/provider.go
utsav14nov: DOne

plugins/providers/gcloudiam/provider.go
utsav14nov: Refactored it to:

resourceRoleMap := make(map[string]*domain.Role)
for _, rc := range pc.Resources {
   for _, ro := range rc.Roles {
      resourceRole := fmt.Sprintf("%s-%s", rc.Type, ro.ID)
      resourceRoleMap[resourceRole] = ro
   }
}
appealResourceRole := fmt.Sprintf("%s-%s", a.Resource.Type, a.Role)
requestedRole := resourceRoleMap[appealResourceRole]

for _, p := range requestedRole.Permissions {
   permission := fmt.Sprint(p)
   if err := client.GrantAccess(a.AccountType, a.AccountID, permission); err != nil {
      if !errors.Is(err, ErrPermissionAlreadyExists) {
         return err
      }
   }
}

plugins/providers/gcloudiam/provider.go
utsav14nov: Same as above

plugins/providers/gcloudiam/provider.go
utsav14nov: Done

plugins/providers/gcloudiam/client.go
utsav14nov: DOne

plugins/providers/gcloudiam/errors.go
```diff
@@ -14,4 +14,6 @@ var (
 	ErrShouldHaveEmptyRoles          = errors.New("gcloud_iam provider should have empty roles")
```
rahmatrhd, line 14: this can be removed since no longer being used everywhere
```suggestion
```

plugins/providers/gcloudiam/provider.go
```diff
@@ -50,6 +52,41 @@ func (p *Provider) GetResources(pc *domain.ProviderConfig) ([]*domain.Resource,
...
+		for _, ro := range rc.Roles {
+			for _, p := range ro.Permissions {
+				permission := fmt.Sprint(p)
+				if _, ok := rolesMap[permission]; !ok {
+					return nil, fmt.Errorf("%v: %v", ErrInvalidProjectRole, permission)
+				}
+			}
+		}
+	}
+
```
rahmatrhd, line 89: this permission validation should be part of `CreateConfig` instead of `GetResources` so that when a user creates/updates a provider config with invalid permissions they will got `ErrInvalidProjectRole`/`ErrInvalidPermission` immediately.

plugins/providers/gcloudiam/provider.go
```diff
@@ -99,12 +152,27 @@ func (p *Provider) RevokeAccess(pc *domain.ProviderConfig, a *domain.Appeal) err
...
+		requestedRole, ok := resourceRoleMap[appealResourceRole]
+		if !ok {
+			return ErrInvalidRole
+		}
+		for _, p := range requestedRole.Permissions {
+			permission := fmt.Sprint(p)
+			if err := client.RevokeAccess(a.AccountType, a.AccountID, permission); err != nil {
+				if !errors.Is(err, ErrPermissionNotFound) {
+					return err
+				}
```
rahmatrhd, line 172: this logic implemented twice in `GrantAccess` and `RevokeAccess`, can you refactor this by moving the logic into a function and reuse it in those both places.

rahmatrhd (changes requested):

plugins/providers/gcloudiam/config.go
```diff
@@ -147,8 +147,8 @@ func (c *Config) validateResourceConfig(resource *domain.ResourceConfig) error {
 	}
 
 	if resource.Roles == nil || len(resource.Roles) == 0 {
-		return nil
+		return ErrRolesShouldNotBeEmpty
 	}
 
-	return ErrShouldHaveEmptyRoles
```
bsushmith, line 153: Can you also validate permissions like how it's done in `bigquery`? https://github.com/odpf/guardian/blob/main/plugins/providers/bigquery/config.go#L169

bsushmith (changes requested):

plugins/providers/gcloudiam/provider.go
mabdh: Hi @utsav14nov @bsushmith , use `%w` to wrap error
```
fmt.Errorf("%w: %v", ErrInvalidProjectRole, permission)
```

plugins/providers/gcloudiam/config.go
```diff
@@ -147,8 +147,8 @@ func (c *Config) validateResourceConfig(resource *domain.ResourceConfig) error {
 	}
 
 	if resource.Roles == nil || len(resource.Roles) == 0 {
```
mabdh, line 149: this is redundant, afaik we just need to check if `len(resource.Roles) == 0`
[reference](https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/builtin/builtin.go;l=161)

plugins/providers/gcloudiam/errors.go
utsav14nov: Done

plugins/providers/gcloudiam/provider.go
utsav14nov: Done

plugins/providers/gcloudiam/provider.go
utsav14nov: Done

plugins/providers/gcloudiam/config.go
utsav14nov: Done

plugins/providers/gcloudiam/config.go
utsav14nov: Done

bsushmith:
@utsav-agarwal overall code test coverage [decreased](https://github.com/odpf/guardian/pull/223#issuecomment-1191340262). Can you check and add test cases if missing for any flow?

plugins/providers/gcloudiam/config.go
```diff
@@ -146,9 +147,41 @@ func (c *Config) validateResourceConfig(resource *domain.ResourceConfig) error {
...
+	rolesMap := make(map[string]*domain.Role)
+	for _, role := range roles {
+		rolesMap[role.ID] = role
+	}
+
+	for _, ro := range resource.Roles {
+		for _, p := range ro.Permissions {
+			permission := fmt.Sprint(p)
+			if _, ok := rolesMap[permission]; !ok {
+				return fmt.Errorf("%v: %v", ErrInvalidProjectRole, permission)
```
rahmatrhd, line 181:
```suggestion
				return fmt.Errorf("%w: %v", ErrInvalidProjectRole, permission)
```

plugins/providers/gcloudiam/provider.go
```diff
@@ -34,6 +36,16 @@ func (p *Provider) CreateConfig(pc *domain.ProviderConfig) error {
 		return err
 	}
 
+	client, err := p.getIamClient(pc)
+	if err != nil {
+		return err
+	}
+
+	r := c.ProviderConfig.Resources[0]
```
rahmatrhd, line 44: even though we will always have one resource and it's also already validated before it goes here, let's still use `for _, r := range c.ProviderConfig.Resources` to avoid confusion and `index out of range error` if there's a change in the future

plugins/providers/gcloudiam/provider.go
```diff
@@ -75,12 +104,19 @@ func (p *Provider) GrantAccess(pc *domain.ProviderConfig, a *domain.Appeal) erro
...
 	if a.Resource.Type == ResourceTypeProject || a.Resource.Type == ResourceTypeOrganization {
-		if err := client.GrantAccess(a.AccountType, a.AccountID, a.Role); err != nil {
-			if errors.Is(err, ErrPermissionAlreadyExists) {
-				return nil
-			}
+		requestedRole, err := p.GetRequestedRoleInAppeal(pc, a)
+		if err != nil {
 			return err
 		}
+		for _, p := range requestedRole.Permissions {
```
rahmatrhd, line 111: after changes in https://github.com/odpf/guardian/pull/220, we shouldn't retrive permissions from `RoleConfig` anymore. Let's use `appeal.Permissions` instead

rahmatrhd (changes requested):

plugins/providers/gcloudiam/provider.go
utsav14nov: Done

plugins/providers/gcloudiam/provider.go
utsav14nov: Done

utsav14nov:
@bsushmith Overall test coverage is fixed.

rahmatrhd (approved):
lgtm

Issue #202, opened 2022-07-05
Bulk appeal access revocation


singhvikash11:
**Summary**
Guardian does not consider the user status like quit, inactive etc in org/company, there is a possibility that users can switch org while they still have active appeals in the system. These accounts pose a huge security threat. When they grow in number, they also become unmanageable for IT administrators. Hence it is essential to periodically check for inactive user accounts and take appropriate actions on them as soon as they're identified.

**Proposed solution**
Guardian appeal-approval workflow takes identity providers(IAM) like Shied, Ldap, and HTTP to fetch user details like name, title, manager, department, etc.  
As a solution we are proposing the following steps:

- Identity providers to have a user status as metadata like `accountStatus=active`
- a scheduled job in Guardian 
    - To fetch all users having active appeals
    - fetch the status of a user using IAM 
    - revoke appeal access if a user is not active

**Additional context**
In the future, we want to trace whether a user is idle for more than a configured duration and will mark a user as inactive and revoke its access.

ravisuhag:
@singhvikash11 Are we expecting that Guardian will expect `accountStatus` key to be there in identity manager? I think this should be kept flexible and there could be mapping that can de be defined to figure out which key to look for. Thoughts?

singhvikash11:
@ravisuhag yes, we'll make it flexible. Users can map `accountStatus` in `IAMSchema`, for now we are not creating any expression to evaluate the status of user.

ravisuhag:
@rahmatrhd @singhvikash11 Based on the discussion on the PR. I have a few questions?

Since IAM is registered in policy, the account status config can change based on each policy. 
- How do we expect the IAM provider to check for it for each policy registered in a generic way? 
- Would dormant access revocation happen at the policy level? 
- Will the job check status config for each policy and then revoke access for users?

singhvikash11:
@ravisuhag IMO, this is as part of the access-request lifecycle in a broader way. Here as a lifecycle of the access request, it filters all the active appeals and kicks-in revocation stage if the user of these appeals becomes `inactive` based on a policy configured.

ravisuhag:
Few thoughts:

1. I think most of the logic should be contained within Job only. We should avoid making changes in core for this apart from introducing status which can be configured in IAM config. 
2. If such actions increase more and more we should think of a way to make jobs generic. On possible way could be. Job can be defined in a yaml config. 

```
action: appeal.revoke 
when: 
on: 
```

rahmatrhd:
[...]

PR #224, opened 2022-07-21
feat: add bulk appeal access revocation

singhvikash11:
Implement [feature](https://github.com/odpf/guardian/issues/202): Bulk appeal access revocation


ravisuhag:
@singhvikash11 I don't see a proton PR for the API contract change. Can you provide a link for same please.

api/handler/v1beta1/appeal.go
```diff
@@ -195,6 +195,44 @@ func (s *GRPCServer) RevokeAppeal(ctx context.Context, req *guardianv1beta1.Revo
...
 
+func (s *GRPCServer) RevokeAppeals(ctx context.Context, req *guardianv1beta1.RevokeAppealsRequest) (*guardianv1beta1.RevokeAppealsResponse, error) {
+	filters := &domain.ListAppealsFilter{}
+	if req.GetAccountId() != nil {
+		filters.AccountID = req.GetAccountId()
+	}
+	filters.Statuses = []string{domain.AppealStatusActive}
+	if req.GetProviderTypes() != nil {
+		filters.ProviderTypes = req.GetProviderTypes()
+	}
```
mabdh, line 206:
```suggestion
	filters.ProviderTypes = req.GetProviderTypes()
```

We can just directly call this without checking nil, getter function will have nil checking inside it

core/appeal/service.go
```diff
@@ -525,6 +531,97 @@ func (s *Service) Revoke(ctx context.Context, id string, actor, reason string) (
 	return revokedAppeal, nil
 }
 
+func (s *Service) BulkRevoke(ctx context.Context, filters *domain.ListAppealsFilter, actor, reason string) ([]*domain.Appeal, error) {
+	result := make([]*domain.Appeal, 0)
+	appeals, err := s.Find(ctx, filters)
```
rahmatrhd, line 536: we can hardcode the list appeals filter.status to be only `active` since we should only revoke active appeals, and also to reduce load to db

rahmatrhd (changes requested):

api/handler/v1beta1/appeal.go
```diff
@@ -195,6 +195,44 @@ func (s *GRPCServer) RevokeAppeal(ctx context.Context, req *guardianv1beta1.Revo
 	}, nil
 }
 
+func (s *GRPCServer) RevokeAppeals(ctx context.Context, req *guardianv1beta1.RevokeAppealsRequest) (*guardianv1beta1.RevokeAppealsResponse, error) {
+	filters := &domain.ListAppealsFilter{}
+	if req.GetAccountId() != nil {
+		filters.AccountID = req.GetAccountId()
```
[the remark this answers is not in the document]
ravisuhag: +1 @singhvikash11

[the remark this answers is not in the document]
singhvikash11: done

api/handler/v1beta1/appeal.go
singhvikash11: done

api/handler/v1beta1/grpc.go
```diff
@@ -75,6 +75,7 @@ type appealService interface {
 	MakeAction(context.Context, domain.ApprovalAction) (*domain.Appeal, error)
 	Cancel(context.Context, string) (*domain.Appeal, error)
 	Revoke(ctx context.Context, id, actor, reason string) (*domain.Appeal, error)
+	BulkRevoke(ctx context.Context, filters *domain.ListAppealsFilter, actor, reason string) ([]*domain.Appeal, error)
```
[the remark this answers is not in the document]
singhvikash11: no, creating a new filter will add duplicate methods like s.repo.Find(listAppealFilter) and s.repo.Find(bulkRevokeFilter). So I am reusing the filter and repo method to filter appeals.

core/appeal/service.go
```diff
@@ -119,6 +119,12 @@ type Service struct {
 	TimeNow func() time.Time
 }
 
+type Key struct {
```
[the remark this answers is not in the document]
singhvikash11: done

api/handler/v1beta1/appeal.go
```diff
@@ -54,7 +54,7 @@ func (s *GRPCServer) ListUserAppeals(ctx context.Context, req *guardianv1beta1.L
 func (s *GRPCServer) ListAppeals(ctx context.Context, req *guardianv1beta1.ListAppealsRequest) (*guardianv1beta1.ListAppealsResponse, error) {
 	filters := &domain.ListAppealsFilter{}
 	if req.GetAccountId() != "" {
-		filters.AccountID = req.GetAccountId()
+		filters.AccountID = []string{req.GetAccountId()}
```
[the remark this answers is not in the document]
singhvikash11: I am trying to do reusability and so use the same filter and underlying repository function

core/appeal/service.go
singhvikash11: done

ravisuhag:
@singhvikash11 Can we please update the docs as well with these changes?

api/handler/v1beta1/appeal.go
```diff
@@ -54,7 +54,7 @@ func (s *GRPCServer) ListUserAppeals(ctx context.Context, req *guardianv1beta1.L
 func (s *GRPCServer) ListAppeals(ctx context.Context, req *guardianv1beta1.ListAppealsRequest) (*guardianv1beta1.ListAppealsResponse, error) {
 	filters := &domain.ListAppealsFilter{}
 	if req.GetAccountId() != "" {
```
mabdh, line 56: The one in the proto file should also `account_ids` right?

api/handler/v1beta1/appeal.go
```diff
@@ -54,7 +54,7 @@ func (s *GRPCServer) ListUserAppeals(ctx context.Context, req *guardianv1beta1.L
...
-		filters.AccountID = req.GetAccountId()
+		filters.AccountIDs = []string{req.GetAccountId()}
 	}
 	if req.GetStatuses() != nil {
```
mabdh, line 59: This also no need to check nil. GetStatuses I think already handle that.

api/handler/v1beta1/appeal.go
mabdh, line 56: Do we really need to check again here?

api/handler/v1beta1/appeal.go
```diff
@@ -195,6 +195,33 @@ func (s *GRPCServer) RevokeAppeal(ctx context.Context, req *guardianv1beta1.Revo
 	}, nil
 }
 
+func (s *GRPCServer) RevokeAppeals(ctx context.Context, req *guardianv1beta1.RevokeAppealsRequest) (*guardianv1beta1.RevokeAppealsResponse, error) {
+	filters := &domain.ListAppealsFilter{}
+	filters.AccountIDs = req.GetAccountIds()
+	filters.ProviderTypes = req.GetProviderTypes()
+	filters.ProviderURNs = req.GetProviderUrns()
+	filters.ResourceTypes = req.GetResourceTypes()
+	filters.ResourceURNs = req.GetResourceUrns()
```
mabdh, line 204: maybe like this more be straightforward
```suggestion
	filters := &domain.ListAppealsFilter{
	    AccountIDs : req.GetAccountIds()
	    ProviderTypes : req.GetProviderTypes()
	    ProviderURNs : req.GetProviderUrns()
	    ResourceTypes : req.GetResourceTypes()
	    ResourceURNs : req.GetResourceUrns()
	}
```

api/handler/v1beta1/appeal.go
```diff
@@ -195,6 +195,33 @@ func (s *GRPCServer) RevokeAppeal(ctx context.Context, req *guardianv1beta1.Revo
...
+
+	reason := req.GetReason()
+
+	appeals, err := s.appealService.BulkRevoke(ctx, filters, domain.SystemActorName, reason)
+	if err != nil {
+		return nil, status.Error(codes.Internal, "failed to revoke appeals in bulk")
+	}
```
mabdh, line 211: Could just pass the reason if it is not being used anywhere
```suggestion
	appeals, err := s.appealService.BulkRevoke(ctx, filters, domain.SystemActorName, req.GetReason())
	if err != nil {
		return nil, status.Error(codes.Internal, "failed to revoke appeals in bulk")
	}
```

api/handler/v1beta1/grpc.go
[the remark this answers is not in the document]
rahmatrhd: instead, we can transform `BulkRevokeFilter` to `domain.ListAppealFilter` in appealService.BulkRevoke as well as adding the appeal status `active` to the list filter. Because it doesn't make sense to have appeal status filter in revoke since we can only revoke active appeals

core/appeal/service.go
```diff
@@ -534,6 +534,98 @@ func (s *Service) Revoke(ctx context.Context, id string, actor, reason string) (
...
+		if resourceAppeals, ok = resourceAppealMap[appeal.ResourceID]; ok {
+			resourceAppeals = append(resourceAppeals, appeal)
+		} else {
+			resourceAppeals = []*domain.Appeal{appeal}
+		}
+		resourceAppealMap[appeal.ResourceID] = resourceAppeals
+	}
+
+	for _, resourceAppeals := range resourceAppealMap {
+		go s.expiredInActiveUserAppeal(ctx, timeLimiter, done, actor, reason, resourceAppeals)
```
mabdh, line 580: Any reason why we need to parallelized here? How is the `totalRequests` that we expected?
For each go routine there is an update task to DB. So this will create mutiple concurrent row updates at the same time.

singhvikash11: this can be 10-50 based on the frequency of job runs.

api/handler/v1beta1/appeal.go
singhvikash11: done

api/handler/v1beta1/grpc.go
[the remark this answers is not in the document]
singhvikash11: @rahmatrhd The [API contract](https://github.com/odpf/proton/blob/main/odpf/guardian/v1beta1/guardian.proto#L437) of bulk revoke doesn't take appeal status as a filter. Can you please suggest the changes.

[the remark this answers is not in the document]
rahmatrhd: > The [API contract](https://github.com/odpf/proton/blob/main/odpf/guardian/v1beta1/guardian.proto#L437) of bulk revoke doesn't take appeal status as a filter. Can you please suggest the changes.

then it's better to have its own filter struct instead of using ListAppealFilter, the code later could be like this 
```go
func (s *service) BulkRevoke(ctx context.Context, filters *domain.RevokeAppealsFilter, actor, reason string) ([]*domain.Appeal, error) {
  appeals, err := s.repository.List(&domain.ListAppealsFilter{
    Status: 'active',
    ResourceURN: filters.ResourceURN,
    ResourceType: filters.ResourceType,
    ...
  })
  ...
}
```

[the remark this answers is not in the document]
singhvikash11: done

ravisuhag:
@singhvikash11 Can you also add a CLI command for this API?

domain/appeal.go
```diff
@@ -126,7 +126,7 @@ type ApprovalAction struct {
 
 type ListAppealsFilter struct {
 	CreatedBy                 string    `mapstructure:"created_by" validate:"omitempty,required"`
-	AccountID                 string    `mapstructure:"account_id" validate:"omitempty,required"`
+	AccountIDs                []string  `mapstructure:"account_ids" validate:"omitempty,required"`
```
rahmatrhd, line 129: this will be a breaking change, we can still keep `AccountID` with a "deprecated" comment and add the new `AccountIDs` to keep it backward compatible. Therefore, no implementation changes should be done in ListAppeals API apart from `AccountIDs` filter addition

rahmatrhd (changes requested):

mabdh (approved):

docs/docs/reference/api.md
```diff
@@ -505,9 +505,9 @@ The request parameters associated with this is API are as follows:
 
 ##### Parameters
 
-| Name           | Located in | Description | Required | Type   |
-| -------------- | ---------- | ----------- | -------- | -------- |
-| account_id     | query      |             | No | string   |
```
rahmatrhd, line 510: let's keep this `account_id` and add deprecation note

core/appeal/service.go
```diff
@@ -556,6 +556,110 @@ func (s *Service) Revoke(ctx context.Context, id string, actor, reason string) (
 	return revokedAppeal, nil
 }
 
+func (s *Service) BulkRevoke(ctx context.Context, filters *domain.RevokeAppealsFilter, actor, reason string) ([]*domain.Appeal, error) {
+	if filters.AccountIDs == nil || len(filters.AccountIDs) == 0 {
+		return nil, fmt.Errorf("account_ids is missing")
```
rahmatrhd, line 561:
```suggestion
		return nil, fmt.Errorf("account_ids is required")
```

internal/store/postgres/appeal_repository_test.go
```diff
@@ -321,9 +321,9 @@ func (s *AppealRepositoryTestSuite) TestFind() {
 			},
 			{
 				filters: &domain.ListAppealsFilter{
-					AccountID: "user@email.com",
```
rahmatrhd, line 324: let's also keep the `AccountID` one in the test case

rahmatrhd (changes requested):

docs/docs/reference/api.md
singhvikash11: done

core/appeal/service.go
singhvikash11: done

internal/store/postgres/appeal_repository_test.go
singhvikash11: done

Issue #194, opened 2022-06-27
Add support for GCS provider


mabdh:
**Summary**
We need to support access management of GCS.

**Proposed solution**

- [x]  Provider configuration for gcs
- [x] GCS client
- [x] GCS resource & access management (TODO: figure out what resources that need to be granted & revoked)
- [x] Documentation

Proposed Provider Config:
```yaml
type: gcs/gcloud_storage
urn: my-google-cloud-storage
allowed_account_types: 
    -user 
    -serviceAccount 
    -group
    -domain
credentials:
  Service_account_key: base64 encoded Service Account Key
  Resource_name: projects/gcs-project-id
resources:
-type: bucket
  policy:
    id: my_bucket_policy
    version: 1
  roles:
    - id: viewer
      name: viewer
      description: ...
      permissions:
        - roles/storage.objectViewer
    - id: owner
      name: OWNER
      description: ...
      permissions:
        - roles/storage.objectCreator
    - id: admin
      name: ADMIN
      description: ...
      permissions: 
         -roles/storage.objectAdmin
-type: object
  policy:
     id: my_object_policy
     version: 1
  roles:
    - id: viewer
      name: View
      description: ...
      permissions:
         - reader
    - id: owner
      name: OWNER
      description: ...
      permissions:
         - owner
```
Resource Config for Bucket 
```yaml
{
  "id": 1,
  "provider_type": "gcs",
  "provider_urn": "my-gcs",
  "type": "bucket",
  "urn": "my-bucket-name",
  "name": "my-bucket-name",
  "details": {
    "foo": "bar"
  }
}
```
Resource Config for Object
```yaml
{
  "id": 1,
  "provider_type": "gcs",
  "provider_urn": "my-gcs",
  "type": "object",
  "urn": "folder/sub-folder/file.txt",
  "name": "file.txt",
  "details": {
    "foo": "bar"
  }
}
```

bsushmith:
> Can we proceed with the plan that every user is assigned a unique bucket

@Chief-Rishab 
Can you elaborate on this `unique` part here? Do you mean a new bucket for each user? 


I'm not too sure about object-level access given your comment. Even with the bulk fetch, there is a high possibility that there are thousands of objects/files and this will increase the table size dramatically with fewer benefits considering the operating issues. Even if we are able to store, the resource list could get too long to deliver a meaningful value for the end user.

Ensuring that onboards onto guardian only those buckets which have a limited(not too many) numbers of objects, is not maintainable. 

[...]

PR #227, opened 2022-07-22
feat: add gcs provider

Chief-Rishab:
Issue linked to this branch-  https://github.com/odpf/guardian/issues/194
Summary
We need to support access management of GCS.

Proposed solution
- [x] Provider configuration for gcs
- [x] GCS client
- [x] GCS resource & access management 
- [x] Documentation
- [x] Tests


docs/docs/providers/gcs.md
```diff
@@ -17,7 +17,38 @@ Access can be given at the bucket level or object level as those allowed to be m
...
+credentials:
+  service_account_key: <base64 encoded service account key json>
+  resource_name: projects/<gcs-project-id>
+  - type: bucket
+    policy:
+      id: my-first-policy
+      version: 1
+    roles:
+      - id: READER
+        name: 'Grants permission to list a bucket contents and read bucket metadata, excluding IAM policies'
```
rahmatrhd, line 32: `name` should be the role's display name, and this can go into `description` field

Chief-Rishab: pushed in the latest commit

plugins/providers/gcs/client.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+
+	"cloud.google.com/go/iam"
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+)
+
+//go:generate mockery --name=GcsClient --exported --with-expecter
+type GcsClient interface {
```
mabdh, line 14:
```suggestion
type GCSClient interface {
```

reference https://github.com/golang/go/wiki/CodeReviewComments#initialisms

plugins/providers/gcs/config.go
```diff
@@ -0,0 +1,195 @@
...
+
+	credentials, err := c.validateCredentials(c.ProviderConfig.Credentials)
+	if err != nil {
+		validationError = append(validationError, err)
+	} else {
+		c.ProviderConfig.Credentials = credentials
+	}
+	ctx := context.TODO()
+	saKey := credentials.ServiceAccountKey
+	fmt.Printf("saKey: %v\n", saKey)
```
mabdh, line 112: maybe you forgot to remove this

plugins/providers/gcs/client.go
Chief-Rishab: Thanks for the reference mas @mabdh. Will update the changes accordingly

plugins/providers/gcs/config.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	var pc Permission
+	if err := mapstructure.Decode(permissionConfig, &pc); err != nil {
+		return nil, err
+	}
+
+	var nameValidation string
+	if resourceType == ResourceTypeBucket {
+		nameValidation = "oneof=roles/storage.admin roles/storage.legacyBucketOwner roles/storage.legacyBucketReader roles/storage.legacyBucketWriter roles/storage.legacyObjectOwner roles/storage.legacyObjectReader roles/storage.objectAdmin roles/storage.objectCreator roles/storage.objectViewer"
```
mabdh, line 188: can we define these validation strings as const?

plugins/providers/gcs/provider.go
```diff
@@ -0,0 +1,251 @@
...
+
+	client, err := p.getGcsClient(creds)
+	if err != nil {
+		return fmt.Errorf("error in getting new client: %w", err)
+	}
+
+	user := a.AccountID
+	userType := a.AccountType
+	identity := fmt.Sprintf("%s:%s", userType, user)
+	ctx := context.TODO()
```
mabdh, line 110: why use `context.TODO()` here?

plugins/providers/gcs/provider.go
```diff
@@ -0,0 +1,251 @@
...
+	if a.Resource.Type == ResourceTypeBucket {
+		b := new(Bucket)
+		if err := b.fromDomain(a.Resource); err != nil {
+			return fmt.Errorf("from Domain func error: %w", err)
+		}
+		for _, p := range permissions {
+			var role iam.RoleName = iam.RoleName(string(p))
```
mabdh, line 117: we can just do this right?
```suggestion
			var role = iam.RoleName(string(p))
```
or
```suggestion
			role := iam.RoleName(string(p))
```

plugins/providers/gcs/config.go
Chief-Rishab: yep, will do it

plugins/providers/gcs/provider.go
```diff
@@ -0,0 +1,251 @@
...
+		return nil, err
+	}
+
+	resourceTypes := []string{}
+	for _, rc := range pc.Resources {
+		resourceTypes = append(resourceTypes, rc.Type)
+	}
+
+	var resources []*domain.Resource
+	ctx := context.Background()
```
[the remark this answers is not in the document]
Chief-Rishab: the GetBucket api required that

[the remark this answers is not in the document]
mabdh: why not forwarding the context from the request?

plugins/providers/gcs/config.go
```diff
@@ -0,0 +1,195 @@
...
+func (c *Config) parseAndValidate() error {
+	if c.valid {
+		return nil
+	}
+
+	validationError := []error{}
+
+	credentials, err := c.validateCredentials(c.ProviderConfig.Credentials)
+	if err != nil {
+		validationError = append(validationError, err)
```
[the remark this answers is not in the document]
Chief-Rishab: validateCredentials and validateResourceConfig are two different functions we are calling, the errors return in this case will have the slice of errors from both of these functions.

plugins/providers/gcs/config.go
Chief-Rishab: updated in the recent commit

plugins/providers/gcs/config.go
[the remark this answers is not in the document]
mabdh: yes I know but why should we not return error at this point? why should we bother creating a storage.NewClient if we return error anyway?
cc @rahmatrhd 


note: please not resolving the discussion if it is not resolved yet

plugins/providers/gcs/provider.go
mabdh: @rahmatrhd can we get the context from request and propagated that here? or should this call is being decoupled from the propagated context?

rahmatrhd: yes ideally it should be there, but in order to do that we need to update the ProviderClient interface which also require to update the other provider plugins methods. I think we can do this in another PR

plugins/providers/gcs/client.go
```diff
@@ -0,0 +1,89 @@
...
+		result = append(result, &Bucket{
+			Name: battrs.Name,
+		})
+	}
+	return result, nil
+}
+
+func (c *gcsClient) GrantBucketAccess(ctx context.Context, b *Bucket, identity string, role iam.RoleName) error {
+
+	bucketName := b.Name
```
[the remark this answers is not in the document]
Chief-Rishab: updated this latest commit, updated the mockery as well

plugins/providers/gcs/provider.go
mabdh: I see got it. I think it is fair enough to pass `context.TODO` in this provider. I believe we need an [issue](https://github.com/odpf/guardian/issues/240) like this later?

plugins/providers/gcs/resource.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	b.Name = r.URN
+	return nil
+}
+
+func (b *Bucket) toDomain() *domain.Resource {
+	return &domain.Resource{
+		Type: ResourceTypeBucket,
+		URN:  b.Name,
```
mabdh, line 25: does URN always equal to the name (or vice versa)?

plugins/providers/gcs/client.go
```diff
@@ -0,0 +1,89 @@
...
+			return nil, err
+		}
+		result = append(result, &Bucket{
+			Name: battrs.Name,
+		})
+	}
+	return result, nil
+}
+
+func (c *gcsClient) GrantBucketAccess(ctx context.Context, b Bucket, identity string, role iam.RoleName) error {
```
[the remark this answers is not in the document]
mabdh: Yes I mean we can add comment in the code so the reader could understand what do we mean by `identity`.

plugins/providers/gcs/resource.go
Chief-Rishab: yes, the resource URN(in guardian) is the actual resource name(in the provider) which the user might want to create an appeal for. Might differ from provider to provider. For a `bucket` resource, @rahmatrhd and I discussed having URN the same as the bucket name. And for the `object`(no longer supported) it was the complete path. 

Reference: the Resource Config for Bucket and Object https://github.com/odpf/guardian/issues/194

plugins/providers/gcs/provider.go
```diff
@@ -0,0 +1,255 @@
...
+			role := iam.RoleName(string(p))
+			if err := client.GrantBucketAccess(ctx, *b, identity, role); err != nil {
+				if errors.Is(err, ErrPermissionAlreadyExists) {
+					return nil
+				}
+				return err
+			}
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
Chief-Rishab: Sure, will update accordingly

plugins/providers/gcs/provider.go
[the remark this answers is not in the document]
Chief-Rishab: this was the same reason as discussed below, the surrounding function hasn't received the context, and we are required to use it further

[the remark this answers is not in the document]
mabdh: Yes, context.TODO i think is more proper here. Can we update it?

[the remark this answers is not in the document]
Chief-Rishab: sure, will update it

plugins/providers/gcs/client.go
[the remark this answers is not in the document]
Chief-Rishab: sure, will do it in the next commit

plugins/providers/gcs/config.go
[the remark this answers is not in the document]
Chief-Rishab: @mabdh I have removed the NewClient initialisation, it wasn't required in the create config function. I just had to check for credentials validity in this function.

plugins/providers/gcs/config.go
```diff
@@ -0,0 +1,194 @@
...
+	valid          bool
+
+	crypto    domain.Crypto
+	validator *validator.Validate
+}
+
+func (c *Config) EncryptCredentials() error {
+	if err := c.parseAndValidate(); err != nil {
+		return err
+	}
```
mabdh, line 42: this won't be needed if we sure the credential is always valid before this `EncrypCredentials` being called right?

plugins/providers/gcs/config.go
```diff
@@ -0,0 +1,194 @@
...
+	if len(validationError) > 0 {
+		errorStrings := []string{}
+		for _, err := range validationError {
+			errorStrings = append(errorStrings, err.Error())
+		}
+		return errors.New(strings.Join(errorStrings, "\n"))
+	}
+
+	c.valid = true
+	return nil
```
mabdh, line 133: this valid boolean is also not needed if this validation is only being done once when we read the config. we can make this function a pure function by not changing the state.

plugins/providers/gcs/provider.go
```diff
@@ -0,0 +1,253 @@
...
+	if err := creds.Decrypt(p.crypto); err != nil {
+		return fmt.Errorf("error in decrypting credentials%w", err)
+	}
+
+	client, err := p.getGCSClient(creds)
+	if err != nil {
+		return fmt.Errorf("error in getting new client: %w", err)
+	}
+
+	identity := fmt.Sprintf("%s:%s", a.AccountType, a.AccountID) // identity is AccountType : AccountID, eg: "serviceAccount:test@email.com"
```
mabdh, line 112: you can put the comment above it to make it easier to read
```suggestion
        // identity is AccountType : AccountID, eg: "serviceAccount:test@email.com"
	identity := fmt.Sprintf("%s:%s", a.AccountType, a.AccountID) 
```

plugins/providers/gcs/provider.go
```diff
@@ -0,0 +1,253 @@
...
+			var role iam.RoleName = iam.RoleName(string(p))
+			if err := client.RevokeBucketAccess(ctx, *b, identity, role); err != nil {
+				if errors.Is(err, ErrPermissionAlreadyExists) {
+					return nil
+				}
+				return err
+			}
+		}
+		return nil
+	}
```
mabdh, line 175: can just pass the context.TODO directly, no need to introduce a new variable if it is only being used in a single place
```suggestion
	if a.Resource.Type == ResourceTypeBucket {
		b := new(Bucket)
		if err := b.fromDomain(a.Resource); err != nil {
			return fmt.Errorf("from Domain func error: %w", err)
		}
		for _, p := range permissions {
			var role iam.RoleName = iam.RoleName(string(p))
			if err := client.RevokeBucketAccess(context.TODO(), *b, identity, role); err != nil {
				if errors.Is(err, ErrPermissionAlreadyExists) {
					return nil
				}
				return err
			}
		}
		return nil
	}
```

plugins/providers/gcs/client.go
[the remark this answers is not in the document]
Chief-Rishab: done this in the last commit

plugins/providers/gcs/config.go
Chief-Rishab: we were using the same in other providers in guardian. Asked this from @rahmatrhd and he told this boolean value will be required if we're calling the parseAndValidate function apart from the create config function. So the entire computation could be avoided. Will look into it once again

plugins/providers/gcs/provider.go
Chief-Rishab: done in https://github.com/odpf/guardian/pull/227/commits/575cacf66a8181ddf2953708d3c7d26f0b98eba5

plugins/providers/gcs/provider.go
Chief-Rishab: done in https://github.com/odpf/guardian/pull/227/commits/575cacf66a8181ddf2953708d3c7d26f0b98eba5

plugins/providers/gcs/config.go
Chief-Rishab: @rahmatrhd should we update all the other providers with this, I checked this function is being called only once in all the providers. Apart form this the valid bool will also be removed since it was being used to cache for this call only

mabdh: @Chief-Rishab we can let the other providers problem to another PR. Each provider might have different behaviour so IMO we can try to keep this provider simpler.

Chief-Rishab: sure, will do it for gcs then

plugins/providers/gcs/config.go
rahmatrhd: @Chief-Rishab need to check again if this provider plugin requires that. In bigquery provider it was needed

plugins/providers/gcs/provider.go
```diff
@@ -0,0 +1,253 @@
...
+	if a.Resource.ProviderType != pc.Type {
+		return ErrProviderTypeMismatch
+	}
+	if a.Resource.ProviderURN != pc.URN {
+		return ErrProviderURNMismatch
+	}
+	return nil
+}
+
+func getPermissions(resourceConfigs []*domain.ResourceConfig, a *domain.Appeal) ([]Permission, error) {
```
rahmatrhd, line 205: let's update this to retrieve `permissions` from appeal instead of resource config

rahmatrhd (changes requested):

plugins/providers/gcs/provider.go
Chief-Rishab: done in https://github.com/odpf/guardian/pull/227/commits/3718c326b7f07d0e578375986ac9e6a39084197d

plugins/providers/gcs/config.go
Chief-Rishab: @mabdh updated the changes in https://github.com/odpf/guardian/pull/227/commits/b70e3ec4aa8e55ff4938dd43d641c1ce007e8bc3

plugins/providers/gcs/config.go
Chief-Rishab: done in https://github.com/odpf/guardian/pull/227/commits/b70e3ec4aa8e55ff4938dd43d641c1ce007e8bc3

plugins/providers/gcs/provider.go
```diff
@@ -202,38 +195,12 @@ func validateProviderConfigAndAppealParams(pc *domain.ProviderConfig, a *domain.
...
-	for _, r := range resourceConfig.Roles {
-		if r.ID == a.Role {
-			role = r
-		}
-	}
-	if role == nil {
-		return nil, ErrInvalidRole
-	}
-
+func getPermissions(a *domain.Appeal) []Permission {
```
mabdh, line 208: is this function specific for gcs or a generic one?

could this go as function receiver in `domain.Appeal`? so calling this would be `a.GetPermission()`
cc @rahmatrhd

plugins/providers/gcs/provider.go
Chief-Rishab: can we close this conversation? @rahmatrhd

plugins/providers/gcs/provider.go
Chief-Rishab: it's a generic function(there in other providers as well), not sure about the second part of the question. @rahmatrhd

plugins/providers/gcs/config.go
Chief-Rishab: @mabdh can we close this conversation?

plugins/providers/gcs/provider.go
Chief-Rishab: This helper function is provider specific, I don't think it requires to be considered as a receiver function in domain.Appeal. @rahmatrhd

rahmatrhd: this is specific to gcs because it cast permission string into `gcs.Permission` type

rahmatrhd (approved):
lgtm

mabdh (approved):

PR #228, opened 2022-07-22
chore(store): replace gorm's automigration with sql migration using go-migrate

rahmatrhd:


internal/store/postgres/store.go
```diff
@@ -30,26 +41,46 @@ func NewStore(c *store.Config) (*Store, error) {
...
+	if err != nil {
+		return err
+	}
+
+	if err := m.Up(); err != nil {
+		if errors.Is(err, migrate.ErrNoChange) {
+			log.Print("no change")
+		} else {
 			return err
 		}
```
mabdh, line 66:
```suggestion
		if errors.Is(err, migrate.ErrNoChange) {
			log.Print("no change")
			return nil
		}
		return err
```

internal/store/postgres/store.go
```diff
@@ -30,26 +41,46 @@ func NewStore(c *store.Config) (*Store, error) {
...
+		return err
+	}
+	m, err := migrate.NewWithSourceInstance("iofs", iofsDriver, toConnectionString(s.config))
+	if err != nil {
+		return err
+	}
+
+	if err := m.Up(); err != nil {
+		if errors.Is(err, migrate.ErrNoChange) {
+			log.Print("no change")
```
bsushmith, line 63: Can you change this log line to be more descriptive?

internal/store/postgres/store.go
rahmatrhd: @mabdh updated it to be like this instead
```go
        if err := m.Up(); errors.Is(err, migrate.ErrNoChange) {
		log.Println("migration schema version is up to date")
		return nil
	} else if err != nil {
		return err
	}
```

mabdh: I think this is more confusing to read compare to 
```
if err := m.Up(); err != nil {
  if errors.Is(err, migrate.ErrNoChange) {
    log.Println("migration schema version is up to date")
    return nil
  }
  return err
}
```

wdyt @rahmatrhd @bsushmith ?

bsushmith: Even this would be fine - 
```go
	err = m.Up()
	if err != nil && errors.Is(err, migrate.ErrNoChange) {
		log.Println("migration schema version is up to date")
		return nil
	}

	return err
```

rahmatrhd: updated as @mabdh's suggestion

mabdh (approved):

rahmatrhd:
@ravisuhag @AkarshSatija please help to approve and merge

bsushmith (approved):

ravisuhag:
@rahmatrhd can you also resolve merge conflicts please?

Issue #229, opened 2022-07-26
APIs for add and remove approver


rahmatrhd:
**Summary**
There is a case where the only approver is unavailable and the approval got blocked due to that. Need a way to add new approver to a pending appeal.

**Proposed solution**
- new API to add an approver
  `POST /appeals/{id}/approvals/{id_or_name}/approvers`
  body:
  ```json
  {
    "email": "new.approver@example.com"
  }
  ```
- new API to remove approver
  `DELETE /appeals/{id}/approvals/{id_or_name}/approvers/{email}`

bsushmith:
@rahmatrhd What would happen when and if the only approver is removed? Is that step - marked as `auto-approved` ...

rahmatrhd:
@bsushmith  if there's only one approver, that approver can't be removed. Added that validation as one of the task  in the PR's checklist

PR #230, opened 2022-07-26
feat(appeal): add APIs to add and remove approvers

rahmatrhd:
closes: #229 

proton PR: https://github.com/odpf/proton/pull/167

Checklist:
- [x] add approver API
- [x] remove approver API
- [x] add approver CLI command
- [x] remove approver CLI command
- [x] validations
  - [x] can add/remove approvers from `pending` and `blocked` approvals only
  - [x] can add/remove approvers from manual approvals only
  - [x] can add/remove approvers from `pending` appeals only
  - [x] can't remove approver if he is the only approver in the approval step
- [x] send notification to the new approver when adding new approver
- [x] audit log


bsushmith:
@rahmatrhd could not find the proton pr. Can you link it here...

api/handler/v1beta1/approval.go
```diff
@@ -77,7 +77,7 @@ func (s *GRPCServer) UpdateApproval(ctx context.Context, req *guardianv1beta1.Up
 			return nil, status.Errorf(codes.InvalidArgument, "unable to process the request: %v", err)
 		case appeal.ErrActionForbidden:
 			return nil, status.Error(codes.PermissionDenied, "permission denied")
-		case appeal.ErrApprovalNameNotFound:
+		case appeal.ErrApprovalNotFound:
 			return nil, status.Errorf(codes.NotFound, "appeal not found: %v", id)
```
bsushmith, line 81: Is this correct?  it is checking for `ErrApprovalNotFound`, but returning back `"appeal not found"`

rahmatrhd:
@bsushmith added the proton PR in the description

core/appeal/service.go
```diff
@@ -525,6 +529,132 @@ func (s *Service) Revoke(ctx context.Context, id string, actor, reason string) (
...
+	if approvalID == "" {
+		return nil, nil, ErrApprovalIDEmptyParam
+	}
+
+	appeal, err := s.repo.GetByID(appealID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("getting appeal details: %w", err)
+	}
+	if appeal.Status != domain.AppealStatusPending {
+		return nil, nil, fmt.Errorf("%w: can't delete approver to appeal with %q status", ErrUnableToAddApprover, appeal.Status)
```
bsushmith, line 641: this check needs to be moved to the `DeleteApprover` method.

core/approval/service.go
```diff
@@ -16,6 +16,8 @@ import (
 type repository interface {
 	BulkInsert([]*domain.Approval) error
 	ListApprovals(*domain.ListApprovalsFilter) ([]*domain.Approval, error)
+	AddApprover(*domain.Approver) error
+	DeleteApprover(approvalID, email string) error
```
bsushmith, line 20: Shouldn't the func signature be the same for these two repository methods?

internal/store/postgres/approval_repository.go
```diff
@@ -112,3 +115,34 @@ func (r *ApprovalRepository) BulkInsert(approvals []*domain.Approval) error {
...
+		return fmt.Errorf("parsing approver: %w", err)
+	}
+
+	result := r.db.Create(m)
+	if result.Error != nil {
+		return fmt.Errorf("inserting new approver: %w", result.Error)
+	}
+
+	newApprover := m.ToDomain()
+	*approver = *newApprover
```
bsushmith, line 131: Is this needed? In the service layer, it is not being used either way.

core/approval/service.go
rahmatrhd: I assume AddApprover is similar to CreateApprover, so that's why I put the struct as a parameter there, and for the deleteApprover, deletion usually only requires the identifier(s)

internal/store/postgres/approval_repository.go
rahmatrhd: yes, that's because our domain representation of approvers ([]string) is different with the model ([]model.Approver). But just like other creation methods, they usually update the passed reference with the inserted ID

bsushmith: Yes. But below is the service layer method which calls this repository method. The updated reference is not used at all in this case. 
```go
func (s *Service) AddApprover(_ context.Context, approvalID, email string) error {
	return s.repo.AddApprover(&domain.Approver{
		ApprovalID: approvalID,
		Email:      email,
	})
}
```

core/appeal/service.go
```diff
@@ -525,6 +529,135 @@ func (s *Service) Revoke(ctx context.Context, id string, actor, reason string) (
...
+		return nil, nil, ErrApprovalIDEmptyParam
+	}
+
+	appeal, err := s.repo.GetByID(appealID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("getting appeal details: %w", err)
+	}
+
+	var approval *domain.Approval
+	for _, a := range appeal.Approvals {
```
mabdh, line 648: this could be move to appeal struct as a receiver function 
```go
func (ap *Appeal) GetApprovalByID(approvalID string) *domain.Approval {
	for _, a := range ap.Approvals {
		if a.ID == approvalID || a.Name == approvalID {
			return a
		}
	}
        return nil
}
```

later you could do
```go
if approval  := appeal.GetApprovalByID(approvalID); approval != nil {
  return appeal, approval, nil
}
return nil, nil, ErrApprovalNotFound
```

core/appeal/service.go
```diff
@@ -525,6 +529,135 @@ func (s *Service) Revoke(ctx context.Context, id string, actor, reason string) (
...
+		return nil, fmt.Errorf("%w: can't add approver to approval with %q status", ErrUnableToAddApprover, approval.Status)
+	}
+
+	// check if approval type is auto
+	// this approach is the quickest way to assume that approval is auto, otherwise need to fetch the policy details and lookup the approval type which takes more time
+	if approval.Status == domain.ApprovalStatusBlocked && (approval.Approvers == nil || len(approval.Approvers) == 0) {
+		// approval is automatic (strategy: auto) that is still on blocked
+		return nil, fmt.Errorf("%w: can't modify approvers for approval with strategy auto", ErrUnableToAddApprover)
+	}
+
```
mabdh, line 555: Why don't use
```go
switch appeal.Status {
```
?

core/approval/service.go
```diff
@@ -126,6 +128,17 @@ func (s *Service) AdvanceApproval(ctx context.Context, appeal *domain.Appeal) er
...
 }
 
+func (s *Service) AddApprover(_ context.Context, approvalID, email string) error {
+	return s.repo.AddApprover(&domain.Approver{
+		ApprovalID: approvalID,
+		Email:      email,
+	})
+}
+
+func (s *Service) DeleteApprover(_ context.Context, approvalID, email string) error {
```
mabdh, line 138: why don't propagate context to DB?
in DB we use `db.WithContext(ctx)` as well

ravisuhag:
@rahmatrhd Can we please update docs as well for these changes?

core/appeal/service.go
rahmatrhd: added switch for the `approval.Status`

core/approval/service.go
rahmatrhd: created an issue for this https://github.com/odpf/guardian/issues/240

internal/store/postgres/approval_repository.go
rahmatrhd: because when creating this in the repository I was not thinking about how it will be used in the service layer or somewhere else. Was just thinking about the responsibility of the repository itself with the given function signature (pass approver struct by reference)

bsushmith:
LGTM. Please get the proton pr merged and update the makefile. @rahmatrhd

bsushmith (approved):

mabdh (approved):

Issue #212, opened 2022-07-12
Appeal request on behalf of other user


singhvikash11:
**Summary**
As a user, they want to raise an appeal for another user in Guardian. This could help the manager/supervisor to raise all access-request for a user and sometimes this access request is for integration account to other systems like `metabase`, `tableau` etc.

**Proposed solution**
Allow the appeal-request flow to get another user's email as input. Will trigger the approval flow for giving another user and send appeal notification to both users(in case it's a non-human account).

singhvikash11:
@rahmatrhd @ravisuhag currently guardian doesn't allow the creation of appeal for another user and this is checked [here](https://github.com/odpf/guardian/blob/main/core/appeal/service.go#L939). 

IMO, this should be part of policy where an org can define they should/shouldn't create an appeal on behalf of another user. The sample policy will look like this:

> id: allow_create_appeal_for_other_user
steps:
> > name: create_appeal_for_other_user_check
description: appeal will be rejected if its creating for other individual
strategy: auto
approve_if: $appeal.account_type == "user" && $appeal.created_by == $appeal.account_id

Guardian will check whether it's creating an appeal for another individual based on created_by == account_id and handle the communication for both users.

@rahmatrhd @ravisuhag wdyt?

ravisuhag:
@singhvikash11 Can you document how it will change the contract of appeal? Are you proposing that `create_appeal_for_other_user_check` this step will only be needed when admins want to disable the appeal creation on behalf of other users?

singhvikash11:
@ravisuhag yes, As an admin they should be able to add this check as part of the approval flow instead of its default check-in guardian appeal service. 

> id: sample_policy
steps:
  > > name: bigquery_resource_check
    description: appeal will be rejected if its for other than bigquery
    strategy: auto
    approve_if: $appeal.resource.provider_type == "bigquery"

  > > name: create_appeal_for_other_user_check
    description: description: appeal will be rejected if its creating for other individual
    strategy: auto
    approve_if: $appeal.account_type == "user" && $appeal.created_by == $appeal.account_id

  > > name: manager_approval
    description: approval from manager
    strategy: manual
    approvers:
       $appeal.creator.manager_email

rahmatrhd:
Instead, we can just have an enable/disable flag in the policy to allow users to create an appeal for another user to make it more straightforward.

Also other things to note:
1. user details in `creator` field right now is based on the `created_by` field. Need to make it based on `account_id` if the appeal is created by a different user
[...]

PR #244, opened 2022-08-03
feat: allow appeal for other user configurable in policy

singhvikash11:
Implement https://github.com/odpf/guardian/issues/212: Appeal request on behalf of another user


core/appeal/service.go
```diff
@@ -289,6 +293,21 @@ func (s *Service) Create(ctx context.Context, appeals []*domain.Appeal, opts ...
 						},
 					},
 				})
+
+				if appeal.AccountType == domain.DefaultAppealAccountType && appeal.AccountID != appeal.CreatedBy {
```
mabdh, line 297: Can we add a comment explaining this? Or is it possible to extract this out as a function? I saw we are doing similar thing in another place.

domain/notifier.go
```diff
@@ -6,14 +6,16 @@ type NotificationMessages struct {
...
 
 const (
-	NotificationTypeExpirationReminder   = "ExpirationReminder"
-	NotificationTypeAppealApproved       = "AppealApproved"
-	NotificationTypeAppealRejected       = "AppealRejected"
-	NotificationTypeAccessRevoked        = "AccessRevoked"
-	NotificationTypeApproverNotification = "ApproverNotification"
+	NotificationTypeExpirationReminder     = "ExpirationReminder"
+	NotificationTypeAppealApproved         = "AppealApproved"
+	NotificationTypeAppealApprovedForOther = "OthersAppealApproved"
```
mabdh, line 15: Maybe better like this? 

```suggestion
	NotificationTypeAppealForOtherApproved = "OthersAppealApproved"
```

internal/store/postgres/migrations/000006_add_appeal_config_to_policies.up.sql
```diff
@@ -1,2 +1,2 @@
 ALTER TABLE "policies"
-    ADD "appeal_config" jsonb;
\ No newline at end of file
+    ADD "appeal_config" jsonb default '{"allow_cross_individual_user":false}';
```
mabdh, line 2: Can we do this changes without having a default json value like this?
Perhaps having a default value in the app logic itself?

domain/policy.go
```diff
@@ -196,6 +196,7 @@ type Policy struct {
 
 type PolicyAppealConfig struct {
 	DurationOptions []AppealDurationOption `json:"duration_options" yaml:"duration_options" validate:"omitempty,min=1,dive"`
+	AllowOnBehalf   bool                   `json:"allow_on_behalf" yaml:"allow_on_behalf"`
```
mabdh, line 199: Is this the same as allow cross individual user? If yes can we just make the naming consisten?

core/appeal/service.go
bsushmith: +1 Let's have both - separating as a function and then comments there to detail why.

internal/store/postgres/migrations/000006_add_appeal_config_to_policies.up.sql
bsushmith: CMIIW but the domain.PolicyAppealConfig is a pointer in Policy; and if empty/default gets written to DB,  It will be `null`. It need not have this hard coded.

core/appeal/service.go
```diff
@@ -438,6 +469,21 @@ func (s *Service) MakeAction(ctx context.Context, approvalAction domain.Approval
...
+					notifications = append(notifications, domain.Notification{
+						User: appeal.AccountID,
+						Message: domain.NotificationMessage{
+							Type: domain.NotificationTypeAppealApprovedForOther,
+							Variables: map[string]interface{}{
+								"resource_name": fmt.Sprintf("%s (%s: %s)", appeal.Resource.Name, appeal.Resource.ProviderType, appeal.Resource.URN),
+								"role":          appeal.Role,
+								"account_id":    appeal.AccountID,
+								"requestor":     appeal.CreatedBy,
```
bsushmith, line 482: Can you also add the `appeal_id` here? If one wants to modify the message in the config and show `appeal_id` also in the notification, it will be helpful.

core/appeal/service.go
singhvikash11: done

internal/store/postgres/migrations/000006_add_appeal_config_to_policies.up.sql
singhvikash11: done

domain/policy.go
singhvikash11: done

core/appeal/service.go
singhvikash11: done

docs/docs/guides/create-policy.md
```diff
@@ -23,6 +23,13 @@ steps:
     strategy: manual
     approvers:
       - john.doe@company.com
+appeal:
+  - duartion_options:
```
mabdh, line 27:
```suggestion
  - duration_options:
```

domain/notifier.go
```diff
@@ -6,11 +6,13 @@ type NotificationMessages struct {
 	AppealRejected       string `mapstructure:"appeal_rejected" default:"Your appeal to {{.resource_name}} with role {{.role}} has been rejected"`
 	AccessRevoked        string `mapstructure:"access_revoked" default:"Your access to {{.resource_name}}} with role {{.role}} has been revoked"`
 	ApproverNotification string `mapstructure:"approver_notification" default:"You have an appeal created by {{.requestor}} requesting access to {{.resource_name}} with role {{.role}}. Appeal ID: {{.appeal_id}}"`
+	OthersAppealApproved string `mapstructure:"other_appeal_approved" default:"Your appeal to {{.resource_name}} with role {{.role}} created by {{.requestor}} has been approved"`
```
mabdh, line 9: Need to be consistent with the variable and key name. `others` or `other`?
```suggestion
	OthersAppealApproved string `mapstructure:"other_appeal_approved" default:"Your appeal to {{.resource_name}} with role {{.role}} created by {{.requestor}} has been approved"`
```

domain/notifier.go
```diff
@@ -6,11 +6,13 @@ type NotificationMessages struct {
...
 }
 
 const (
 	NotificationTypeExpirationReminder   = "ExpirationReminder"
 	NotificationTypeAppealApproved       = "AppealApproved"
+	NotificationTypeAppealForOnBehalf    = "OnBehalfAppealApproved"
```
mabdh, line 15:
```suggestion
	NotificationTypeOnBehalfAppealApproved    = "OnBehalfAppealApproved"
```

mabdh: I think it is better to have consistent naming for const var and string on this @bsushmith @singhvikash11

bsushmith: Yes @mabdh. This is the updated line as per the suggestion1. Did not commit the suggestion through github UI., as there was one other place it required renaming. 

```go
	NotificationTypeOnBehalfAppealApproved = "OnBehalfAppealApproved"

```

mabdh (approved):

core/appeal/service.go
```diff
@@ -335,6 +360,18 @@ func validateAppealDurationConfig(appeal *domain.Appeal, policy *domain.Policy)
 	return fmt.Errorf("%w: %s", ErrOptionsDurationNotFound, appeal.Options.Duration)
 }
 
+func validateAppealOnBehalf(a *domain.Appeal, policy *domain.Policy) error {
+	if a.AccountType == domain.DefaultAppealAccountType {
+		if policy.AppealConfig != nil && policy.AppealConfig.AllowOnBehalf == true {
```
rahmatrhd, line 365:
```suggestion
		if policy.AppealConfig != nil && policy.AppealConfig.AllowOnBehalf {
```

rahmatrhd (approved):
lgtm, only small suggestion

PR #247, opened 2022-08-04
feat(grant): import pre-existing access from provider

### LINKED ISSUE #206: Import pre-existing access from providers
## Requirements
- Guardian to collect all pre-existing access from each resource in the provider
- User (admin) to be able to revoke that imported access if needed

### 1. Fetching access
- ~~Option 1: add a flag in the provider config to import access~~
**Pros** : Under the assumption that after the provider is onboarded to guardian there will be no access created outside guardian, this approach would be just sufficient (simpler process)
**Cons** : Will only run once when the provider is just created
- Option 2: provide an API endpoint to trigger import access
**Pros** : Can be triggered at any time
**Cons** : Need to be triggered manually
- ~~Option 3: regularly collect existing access with jobs~~
**Pros** : Continuously control those access created outside guardian
**Cons** : Might increase the scope to the `Track access drift` feature

### 2. Creating the appeals
The appeal going to be like this:
```jsonc
{
  "id": "", // auto-generated
  "resource_id": "", // added by guardian
  "policy_id": null,
  "policy_version": 0,
  "status": "active",
  "account_type": "", // imported from provider
  "account_id": "", // imported from provider
  "role": "custom", // TODO: need to find a way to map the existing/imported permissions with the roles user defined in the provider
  "options": {},
  "resource": {}, // added by guardian,
  "approvals": [], // depends on the policy
  "details": {},
  "created_by": "", // TODO: for "user" account type, we can use that as the value here, but can't for serviceAccount or other account types
  "creator": null, // depends on the policy. Might be empty because iam config defined in the policy

  // new field(s):
  "imported": true, // imported flag to differentiate with normal (user-created) appeal
  "permissions": [] // https://github.com/odpf/guardian/issues/205 
}
```

#### Policy
Policy is going to be `null` since there's no approval flow for the imported access
#### Account Types
We are going to import pre-existing access for account types that are defined in the `allowed_account_types` field in the provider config only.
#### Role
Assuming [this bug](https://github.com/odpf/guardian/issues/205) has been resolved,
- **Case 1: the imported access are predefined roles in the Provider Config**
  1. Suppose in the provider config we have defined a role named `bq-admin` which has two permissions: `roles/bigquery.dataViewer` and `roles/bigquery.dataEditor`. 
  2. In case a user has following access granted outside Guardian: `roles/bigquery.dataViewer` and `roles/bigquery.dataEditor`
  3. Those access will be imported and mapped as an appeal with role:`bq-admin` only. Note that it will also have `"permissions": ["roles/bigquery.dataViewer", "roles/bigquery.dataEditor"]` in the appeal object
- **Case 2: the imported access don't have permissions defined in the provider config**
  1. A user has `roles/bigquery.metadataViewer` access in the bigquery, but that permission was not defined in the provider config.
  2. In that case, each access will be mapped into a single appeal with `"role" = "roles/bigquery.metadataViewer"` and `"permissions" = ["roles/bigquery.metadataViewer"]`

rahmatrhd:
closes #206 
proton PR: https://github.com/odpf/proton/pull/197


plugins/providers/client.go
```diff
@@ -10,6 +14,7 @@ type Client interface {
 	RevokeAccess(*domain.ProviderConfig, domain.Grant) error
 	GetRoles(pc *domain.ProviderConfig, resourceType string) ([]*domain.Role, error)
 	GetAccountTypes() []string
+	ListAccess(context.Context, domain.ProviderConfig) (domain.ResourceAccess, error)
```
ravisuhag, line 17: @rahmatrhd As we were discussing, should this provider accept list of resources for which to return access list?

mabdh:
@rahmatrhd is this still in draft?

core/grant/service.go
```diff
@@ -257,3 +268,102 @@ func (s *Service) expiredInActiveUserAccess(ctx context.Context, timeLimiter cha
...
+			grantsMap[g.Resource.URN][userSignature] = map[string]*domain.Grant{}
+		}
+
+		grantsMap[g.Resource.URN][userSignature][g.Role] = &g
+	}
+
+	importedGrants := []*domain.Grant{}
+	for rURN, access := range resourceAccess {
+		for _, ae := range access {
+			userSignature := fmt.Sprintf("%s:%s", ae.AccountType, ae.AccountID)
```
mabdh, line 338: can this be moved to the access itself so calling it would be
```
userSignature := ae.UserSignature()
```

core/grant/service.go
```diff
@@ -257,3 +268,102 @@ func (s *Service) expiredInActiveUserAccess(ctx context.Context, timeLimiter cha
...
+			if grantsMap[rURN] != nil &&
+				grantsMap[rURN][userSignature] != nil &&
+				grantsMap[rURN][userSignature][ae.Permission] != nil {
+				continue // access already registered
+			}
+
+			r := resourcesMap[rURN]
```
mabdh, line 345: this could be done earlier under `for rURN, access := range resourceAccess {` loop

domain/provider.go
```diff
@@ -73,3 +80,13 @@ type ProviderType struct {
...
+
+type AccessEntry struct {
+	Resource    string
+	AccountID   string
+	AccountType string
+	Permission  string
+}
+
+// ResourceAccess is list of UserAccess grouped by resource identifier
+type ResourceAccess map[string][]AccessEntry
```
mabdh, line 92: `MapResourceAccess` might be easier to understand

plugins/providers/bigquery/provider.go
```diff
@@ -3,28 +3,47 @@ package bigquery
...
+	GrantDatasetAccess(ctx context.Context, d *Dataset, user, role string) error
+	RevokeDatasetAccess(ctx context.Context, d *Dataset, user, role string) error
+	GrantTableAccess(ctx context.Context, t *Table, accountType, accountID, role string) error
+	RevokeTableAccess(ctx context.Context, t *Table, accountType, accountID, role string) error
+	ResolveDatasetRole(role string) (bq.AccessRole, error)
+	ListAccess(ctx context.Context, resources []*domain.Resource) (domain.ResourceAccess, error)
+}
+
+//go:generate mockery --name=crypto --exported --with-expecter
+type crypto interface {
```
mabdh, line 28: perhaps `Encryptor` defines it better for interface (although it might do encryption and depcryption)

https://go.dev/doc/effective_go#interface-names

core/grant/service.go
rahmatrhd: added a function `getAccountSignature` in the package instead because it's only utilised specific in grant service logic

plugins/providers/bigquery/provider.go
rahmatrhd: renamed it to encryptor in the bigquery package

mabdh (approved):

Issue #245, opened 2022-08-04
Ability to configure db connection settings


bsushmith:
**Summary**
Ability to configure db connection settings while connecting to the database.

database/sql already supports some of these - 
https://pkg.go.dev/database/sql#DB.SetConnMaxLifetime
https://pkg.go.dev/database/sql#DB.SetMaxIdleConns
https://pkg.go.dev/database/sql#DB.SetMaxOpenConns
https://pkg.go.dev/database/sql#DB.SetMaxOpenConns

ravisuhag:
This will be covered as part of removing gorm. #241

PR #254, opened 2022-08-11
feat: add db connection configuration

singhvikash11:
Address [connection setting](https://github.com/odpf/guardian/issues/245)

### DIFF
--- a/internal/store/config.go
+++ b/internal/store/config.go
@@ -1,12 +1,18 @@
 package store
 
+import "time"
+
 // Config for database connection
 type Config struct {
-	Host     string `mapstructure:"host" default:"localhost"`
-	User     string `mapstructure:"user" default:"postgres"`
-	Password string `mapstructure:"password" default:""`
-	Name     string `mapstructure:"name" default:"postgres"`
-	Port     string `mapstructure:"port" default:"5432"`
-	SslMode  string `mapstructure:"sslmode" default:"disable"`
-	LogLevel string `mapstructure:"log_level" default:"info"`
+	Host            string        `mapstructure:"host" default:"localhost"`
+	User            string        `mapstructure:"user" default:"postgres"`
+	Password        string        `mapstructure:"password" default:""`
+	Name            string        `mapstructure:"name" default:"postgres"`
+	Port            string        `mapstructure:"port" default:"5432"`
+	SslMode         string        `mapstructure:"sslmode" default:"disable"`
+	MaxIdleConns    int           `yaml:"max_idle_conns" mapstructure:"max_idle_conns" default:"3"`
+	MaxOpenConns    int           `yaml:"max_open_conns" mapstructure:"max_open_conns" default:"10"`
+	ConnMaxLifeTime time.Duration `yaml:"conn_max_life_time" mapstructure:"conn_max_life_time" default:"10ms"`
+	MaxIdleLifeTime time.Duration `yaml:"idle_max_life_time" mapstructure:"idle_max_life_time" default:"100ms"`
+	LogLevel        string        `mapstructure:"log_level" default:"info"`
 }
--- a/internal/store/postgres/store.go
+++ b/internal/store/postgres/store.go
@@ -4,6 +4,7 @@ import (
 	"embed"
 	"errors"
 	"fmt"
+	pg "gorm.io/driver/postgres"
 	"log"
 	"net"
 	"net/url"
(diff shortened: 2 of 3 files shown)


internal/store/postgres/store.go
```diff
@@ -41,6 +41,15 @@ func NewStore(c *store.Config) (*Store, error) {
 		log.Panic(err)
 	}
 
+	db, err := gormDB.DB()
+	if err != nil {
+		log.Panic(err)
```
rahmatrhd, line 46:
```suggestion
		return nil, err
```

internal/store/config.go
```diff
@@ -1,12 +1,18 @@
...
+	Host            string        `mapstructure:"host" default:"localhost"`
+	User            string        `mapstructure:"user" default:"postgres"`
+	Password        string        `mapstructure:"password" default:""`
+	Name            string        `mapstructure:"name" default:"postgres"`
+	Port            string        `mapstructure:"port" default:"5432"`
+	SslMode         string        `mapstructure:"sslmode" default:"disable"`
+	MaxIdleConns    int           `yaml:"max_idle_conns" mapstructure:"max_idle_conns" default:"3"`
+	MaxOpenConns    int           `yaml:"max_open_conns" mapstructure:"max_open_conns" default:"10"`
+	ConnMaxLifeTime time.Duration `yaml:"conn_max_life_time" mapstructure:"conn_max_life_time" default:"10ms"`
+	MaxIdleLifeTime time.Duration `yaml:"idle_max_life_time" mapstructure:"idle_max_life_time" default:"100ms"`
```
rahmatrhd, line 16: how do we come up with these default values?

internal/store/postgres/store.go
```diff
@@ -41,6 +41,15 @@ func NewStore(c *store.Config) (*Store, error) {
...
+	}
+	db.SetConnMaxLifetime(c.ConnMaxLifeTime)
+	db.SetMaxIdleConns(c.MaxIdleConns)
+	db.SetMaxOpenConns(c.MaxOpenConns)
+	db.SetConnMaxIdleTime(c.MaxIdleLifeTime)
```
rahmatrhd, line 51: let's add conditions to check if each of these configs doesn't have empty/nil value only then we will configure

rahmatrhd (changes requested):

internal/store/postgres/store.go
bsushmith: @rahmatrhd Is a check needed? Wouldn't the default values configured automatically apply here?

internal/store/config.go
bsushmith: I think these are some sensible defaults that are generally good to have and not based on any specific reasoning. Please CMIIW @singhvikash11

internal/store/postgres/store.go
rahmatrhd: I was thinking if user doesn't explicitly define these additional the config, then let's also not set their values. With this we can also skip adding default values.

bsushmith: It's generally a good practice to set some default DB connection parameters, which the user can override later if they need to. Not setting any is not a great idea. 😅

rahmatrhd: I see, I think we can keep it then 👍

bsushmith:
@singhvikash11 these changes are present in the latest version of salt. Can we bump up the salt version and use that?

internal/store/config.go
```diff
@@ -1,12 +1,18 @@
...
-	LogLevel string `mapstructure:"log_level" default:"info"`
+	Host            string        `mapstructure:"host" default:"localhost"`
+	User            string        `mapstructure:"user" default:"postgres"`
+	Password        string        `mapstructure:"password" default:""`
+	Name            string        `mapstructure:"name" default:"postgres"`
+	Port            string        `mapstructure:"port" default:"5432"`
+	SslMode         string        `mapstructure:"sslmode" default:"disable"`
+	MaxIdleConns    int           `yaml:"max_idle_conns" mapstructure:"max_idle_conns" default:"3"`
+	MaxOpenConns    int           `yaml:"max_open_conns" mapstructure:"max_open_conns" default:"10"`
+	ConnMaxLifeTime time.Duration `yaml:"conn_max_life_time" mapstructure:"conn_max_life_time" default:"10ms"`
```
mabdh, line 15: For the clarity better to add the unit itself.

```suggestion
	ConnMaxLifeTimeMS time.Duration `yaml:"conn_max_life_time_ms" mapstructure:"conn_max_life_time" default:"10ms"`
```

internal/store/config.go
```diff
@@ -1,12 +1,18 @@
 package store
```
mabdh, line 2: we are moving from gorm and I think we suppose to use `salt/db`

ravisuhag:
@singhvikash11 +1, we plan to remove Gorm. So salt/db package will be used for this. If any of the missing setting is not there then lets raise PR for that in salt.

PR #272, opened 2022-08-30
fix: fixes for new access entity

rahmatrhd:
- [x] add order_by query params in list accesses
- [x] treat appeal status 'approved' the same as 'active' in filters for client forward compatibility
- [x] rename access to grant

proton PR: https://github.com/odpf/proton/pull/171


rahmatrhd:
@mabdh reviewing each commit would help because the last commit contains so many changes for renaming access -> grant 😬

ravisuhag (approved):

PR #280, opened 2022-09-02
fix: rename mismatched function name

### LINKED ISSUE #279: get policy preferences api does not work
**Description**
get policy preferences [API](https://github.com/odpf/proton/blob/main/odpf/guardian/v1beta1/guardian.proto#L95) does not work and throws the below error -

```json
{
    "code": 12,
    "message": "method GetPolicyPreferences not implemented"
}
```

This happens due to the function name mismatch between the generated protobuf files and the actual function defined in the handler code of guardian.

protobuf generated code expects - `GetPolicyPreferences`
guardian handler code expects - `GetPolicyAppealConfig`

bsushmith:
Closes #279


rahmatrhd (changes requested):
can you also update the proton commit in the Makefile accordingly?

bsushmith:
This doesn't need a proton change... as this is a name issue in guardian code specifically. proto is fine.
I've run the `make proto` without doing anything., but probably due to proto gen new version - the changes are seen in the proto generated files.

PR #289, opened 2022-09-26
test: implement dockertest for repositories

### LINKED ISSUE #242: Testing store with dockertest postgres instance
**Summary**
Guardian is currently using `go-sqlmock` to test its store layer and  there might be some specific cases that this approach does not cover since we are not testing the real case where a query is running and being evaluated. This could left some edge cases and introduce unwanted bugs.

**Proposed solution**
Use dockertest with a real postgres instance to test store layer.

rahmatrhd:
closes #242


internal/store/postgres/resource_repository_test.go
```diff
@@ -162,7 +162,7 @@ func (s *ResourceRepositoryTestSuite) TestFind() {
 				actualResult, actualError := s.repository.Find(tc.filters)
 
 				s.NoError(actualError)
-				if diff := cmp.Diff(tc.expectedResult, actualResult, cmpopts.EquateApproxTime(time.Nanosecond)); diff != "" {
+				if diff := cmp.Diff(tc.expectedResult, actualResult, cmpopts.EquateApproxTime(time.Microsecond)); diff != "" {
```
rahmatrhd, line 165: so, is it postgres not storing the microsecond? or `cmpopts.EquateApproxTime(time.Microsecond)` is the way to ignore the nanosecond?

mabdh: @rahmatrhd so golang `time.Now` is within nano second precision and postgres uses micro second precision.
above is to ignore up to microsecond difference

internal/store/postgres/policy_repository_test.go
```diff
@@ -1,257 +1,177 @@
...
-					WillReturnRows(sqlmock.NewRows(s.rows).AddRow(expectedRowValues...))
-
-				_, actualError := s.repository.GetOne(tc.expectedID, tc.expectedVersion)
-
-				s.Nil(actualError)
-				s.dbmock.ExpectationsWereMet()
-			})
-		}
-	})
+	//TODO: fix test case
```
mabdh, line 121: @rahmatrhd are we missing this or we don't need this?

internal/store/postgres/grant_repository_test.go
```diff
@@ -2,203 +2,267 @@ package postgres_test
...
+	}
+
+	err = purgeTestDocker(s.pool, s.resource)
+	if err != nil {
+		s.T().Fatal(err)
 	}
 }
 
 func (s *GrantRepositoryTestSuite) TestList() {
+	expDate := time.Now().Truncate(time.Nanosecond)
```
mabdh, line 117: I think this truncating won't be needed anymore

bsushmith (approved):

PR #294, opened 2022-09-27
fix: use created_at, updated_at from appeals otherwise grants table loses the temporal order of access

singhvikash11:
fix: use created_at, updated_at from appeals otherwise grants table loses the temporal order of access

### DIFF
--- a/internal/store/postgres/migrations/000007_create_grants_table.up.sql
+++ b/internal/store/postgres/migrations/000007_create_grants_table.up.sql
@@ -55,8 +55,8 @@ SELECT
   "revoked_at",
   "revoke_reason",
   "created_by",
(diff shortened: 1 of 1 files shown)


rahmatrhd:
I think create/update history can still be referenced from the appeal by using `appeal_id` reference in grant

singhvikash11:
@rahmatrhd we could fetch temporal-order by joining with the appeal table however grants table will have these timestamps populated in case of access is approved for new records. So we need two different queries to find temporal order, one for records before migration and the other after migration.

mabdh:
@singhvikash11 is there any Github issue related with this? If yes can you please link it? If not, can you please create one and explain the context about the bug/changes?

PR #301, opened 2022-10-03
feat: track drift access

### LINKED ISSUE #173: Unauthorised access tracking
**Problem?**
Guardian approval flow with policies makes sure only approved users get access to resources. But if someone was authorized to access directly through the provider platform e.g users are given access to Metabase through its admin portal. Guardian does not have a way to track if access is given to the provider resources which are not being tracked or approved through guardian access workflows.

**Solution**
- Periodically check authorized users at the provider level and track unauthorized access to resources.
- Provide reports and alerts when unauthorized access is detected.
- Raised alert severity can be based on the privacy level of resources.

rahmatrhd:
closes #173 

changes:
- [x] add `source` field to grant
   - grants that created through appeals will have value `source=appeal`
   - imported grants will be `source=import`
- [x] add `status_in_provider` field to track actual access status
- [x] import access from provider and update existing grants' `status_in_provider` accordingly*
- [x] rename `created_by` to `owner`

*) grants <> provider access comparison strategy:
During import, we will get a list of `user`-`permission` access entries, while a `grant` also contains `user` but the `permissions` can be multiple (determined by provider role). So, to be able to map the `grant` to the right access entires, we reduce the access entries based on the `permissions` configured in the role config (provider config).Example:

role config:

id | permissions
--- | ---
viewer | `roles/bigquery.dataViewer`
owner | `roles/bigquery.dataOwner`
admin | `roles/bigquery.dataOwner`, `roles/bigquery.jobUser`

user's active grants:
grant | role | permissions | status | status_in_provider
--- | --- | --- | --- | ---
1 | viewer | `roles/bigquery.dataViewer` | active | active

user's provider access entries on import:
accessEntry | permission
--- | ---
1 | `roles/bigquery.jobUser`
2 | `roles/bigquery.dataOwner`

result:
1. based on accessEntry#1, accessEntry#2, and role config `admin`, we will create a new grant for the user with role `admin`
2. based on the grant#1, user is supposed to have an access entry with `roles/bigquery.dataViewer` permission but it's not. in this case, grant#1 `status_in_provider` will be updated to `inactive`

note: this PR still contains changes from #247


mabdh:
@rahmatrhd some conflicts with the main branch

api/handler/v1beta1/provider.go
```diff
@@ -146,3 +147,35 @@ func (s *GRPCServer) ListRoles(ctx context.Context, req *guardianv1beta1.ListRol
 		Roles: roleProtos,
 	}, nil
 }
+
+func (s *GRPCServer) ImportAccess(ctx context.Context, req *guardianv1beta1.ImportAccessRequest) (*guardianv1beta1.ImportAccessResponse, error) {
+	grants, err := s.grantService.ImportAccess(ctx, grant.ImportAccessCriteria{
+		ProviderID:   req.GetId(),
+		ResourceIDs:  req.GetResourceIds(),
+		ResouceTypes: req.GetResourceTypes(),
+		ResourceURNs: req.GetResourceTypes(),
```
mabdh, line 156: typo ?
```suggestion
		ResourceURNs: req.GetResourceUrns(),
```

internal/store/postgres/migrations/000009_add_grant_source_column.down.sql
```diff
@@ -0,0 +1,18 @@
...
+
+ALTER TABLE
+  "grants"
+ADD
+  COLUMN IF NOT EXISTS "created_by" text;
+
+UPDATE
+  "grants"
+SET
+  "created_by" = "owner";
```
mabdh, line 11: are we renaming column here? or set the column `created_by` value for all rows to  "`owner`".

internal/store/postgres/migrations/000009_add_grant_source_column.up.sql
```diff
@@ -0,0 +1,22 @@
...
+  COLUMN IF NOT EXISTS "status_in_provider" text,
+ADD
+  COLUMN IF NOT EXISTS "owner" text;
+
+UPDATE
+  "grants"
+SET
+  "source" = 'appeal',
+  "status_in_provider" = "status",
+  "owner" = "created_by";
```
mabdh, line 17: same with this, is this for renaming column or setting the specific column with specific value?

core/grant/service.go
```diff
@@ -257,3 +270,176 @@ func (s *Service) expiredInActiveUserAccess(ctx context.Context, timeLimiter cha
...
+		ProviderType: p.Type,
+		ProviderURN:  p.URN,
+	}
+	if criteria.ResourceIDs != nil {
+		listResourcesFilter.IDs = criteria.ResourceIDs
+	} else {
+		listResourcesFilter.ResourceTypes = criteria.ResouceTypes
+		listResourcesFilter.ResourceURNs = criteria.ResourceURNs
+	}
+	resources, err := s.resourceService.Find(ctx, listResourcesFilter)
```
mabdh, line 297: Is it possible to just have a List resource that return resources that has populated provider struct?
So we are just doing 1 DB call to get all resources.
I assume number of resources per provider is not that many.

core/grant/service.go
```diff
@@ -257,3 +270,176 @@ func (s *Service) expiredInActiveUserAccess(ctx context.Context, timeLimiter cha
...
+	if err != nil {
+		return nil, fmt.Errorf("getting resources: %w", err)
+	}
+
+	_, resourceAccess, err := s.providerService.ListAccess(ctx, criteria.ProviderID, resources)
```
mabdh, line 302: This `ListAccess` in provider service has 1 DB call to get provider `GetByID`. I don't think it is necessary because we have called `GetByID` earlier in this function.
I think we can have a `ListAccess` that does not depend on DB call, maybe try to just have a `GetClient` exposed or having a `ListAccess` function that accept provider type argument instead of ID.

core/grant/service.go
```diff
@@ -257,3 +270,176 @@ func (s *Service) expiredInActiveUserAccess(ctx context.Context, timeLimiter cha
...
+		for _, v2 := range v {
+			for _, g := range v2 {
+				g.StatusInProvider = domain.GrantStatusInactive
+				providerInactiveGrants = append(providerInactiveGrants, g)
+			}
+		}
+	}
+	if err := s.repo.BulkUpsert(ctx, providerInactiveGrants); err != nil {
+		return nil, fmt.Errorf("updating grants provider status: %w", err)
+	}
```
mabdh, line 388: Can you please put a comment explaining what this is doing?
Why do we need to separate the imported and the inactive one? Can we just make a DB call for all of them?

core/provider/service.go
```diff
@@ -353,16 +353,41 @@ func (s *Service) Delete(ctx context.Context, id string) error {
 	return nil
 }
 
+func (s *Service) ListAccess(ctx context.Context, providerID string, resources []*domain.Resource) (*domain.Provider, domain.MapResourceAccess, error) {
```
mabdh, line 356: This function would just accept `provider type` and the `GetByID` could be done before the caller call this function. No need to have DB dependency for this.

domain/provider.go
```diff
@@ -73,3 +96,26 @@ type ProviderType struct {
 	Name          string   `json:"name" yaml:"name"`
 	ResourceTypes []string `json:"resource_types" yaml:"resource_types"`
 }
+
+type AccessEntry struct {
+	Resource    string
```
mabdh, line 101: This one resource as in resource id or resource urn?

internal/store/postgres/migrations/000009_add_grant_source_column.down.sql
rahmatrhd: Yes, we are renaming column `created_by` to `owner`, and in this `down` SQL script, setting the value of `created_by` to `"grants"."owner"`

internal/store/postgres/migrations/000009_add_grant_source_column.up.sql
rahmatrhd: yes renaming column, and setting `"grants"."owner"` to `"grants"."created_by"`

mabdh: And for the `"source" = 'appeal',`, is the single quote `'` there is mandatory or could we make it the same with the others and use `"` double quotes?

rahmatrhd: single quotes are for string/text value, so the value is going to be exactly "appeal". And the double quotes are for referring to the column name, we are setting the value of `"grants"."owner"` to be the current value of `"grants"."created_by"`

core/grant/service.go
rahmatrhd: changed `provider.ListAccess` to receive `provider` struct, however it's still need to be part of provider service so it can be used by the other service (in this case grant service)

core/grant/service.go
rahmatrhd: imported grants are fetched from the provider and `inactiveGrants` are the remaining active grants that no longer have access entry in the provider, therefore we marking them as inactive

domain/provider.go
rahmatrhd: this is resourceURN

rahmatrhd: just realized that this field is not being used everywhere, removing it

core/grant/service.go
rahmatrhd: based on the way we are using it right now, this service is called in `POST /providers/:provider_id/import-access` API, so IMO it's still fair to fetch the provider details first, and then the resources

mabdh (approved):

PR #305, opened 2022-10-13
fix: allow 0h as permanent duration option

singhvikash11:


bsushmith (approved):

domain/appeal.go
```diff
@@ -128,8 +128,12 @@ func (a Appeal) ToGrant() (*Grant, error) {
 		if err != nil {
 			return nil, fmt.Errorf("parsing duration %q: %w", a.Options.Duration, err)
 		}
-		expDate := time.Now().Add(duration)
-		grant.ExpirationDate = &expDate
+		if duration == 0*time.Second {
```
mabdh, line 131: do we need the `time.Second` ? or `if duration == 0 {` is enough?

singhvikash11: @mabdh thanks, done

rahmatrhd (approved):

Issue #298, opened 2022-09-29
incremental update fetch resources provider wise


singhvikash11:
**Description**
Guardian fetches all resources once for all providers and updates resources in bulk at the end. Now if there is some data issue with one provider or one resource then the job will not able to update any resources until the data issue fixed

**To Reproduce**
We accidentally change the urn value of a provider but the resources update point to `config-> urn` inside that provider and it starts failing with a foreign key constraint error since the urn's value in the config of a provider does not exist as an urn column in provider table. However, it's very difficult to find from the error log since it prints all the SQL inserts values in batch

**Expected behaviour**
It should update the resources batch-wise so that an incorrect provider will not affect others.

PR #307, opened 2022-10-14
fix: fetch provider resources incrementally 

singhvikash11:
closes : #298

### DIFF
--- a/core/provider/service.go
+++ b/core/provider/service.go
@@ -215,8 +215,7 @@ func (s *Service) FetchResources(ctx context.Context) error {
 			"count", len(resources),
 		)
 		resources = append(resources, res...)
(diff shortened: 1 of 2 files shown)


core/provider/service.go
```diff
@@ -214,9 +215,17 @@ func (s *Service) FetchResources(ctx context.Context) error {
 			"count", len(resources),
 		)
 		resources = append(resources, res...)
+		err = s.resourceService.BulkUpsert(ctx, resources)
+		if err != nil {
```
rahmatrhd, line 219:
```suggestion
		if err := s.resourceService.BulkUpsert(ctx, resources); err != nil {
```

rahmatrhd (changes requested):

rahmatrhd (approved):

PR #317, opened 2022-11-03
feat: allow additional roles for BQ provider

singhvikash11:


plugins/providers/bigquery/client.go
```diff
@@ -327,6 +306,34 @@ func (c *bigQueryClient) ListAccess(ctx context.Context, resources []*domain.Res
 	return access, nil
 }
 
+func (c *bigQueryClient) getGrantAbleRoleForDataset() ([]bq.AccessRole, error) {
+	var roles = []bq.AccessRole{bq.OwnerRole, bq.WriterRole, bq.ReaderRole}
+	var resourceName string
+	ctx := context.Background()
+	datasetIterator := c.client.Datasets(ctx)
+	for {
+		dataset, err := datasetIterator.Next()
```
bsushmith, line 315: Would this iterate over each dataset and their roles in a given project while validating the given role in the config?

If yes... this iteration will take too much time - if a given project has many datasets.

singhvikash11: nope, it'll iterate only one dataset and break, There is no way to add custom role on bigquery dataset so every dataset will have same sort of roles

plugins/providers/bigquery/client.go
```diff
@@ -327,6 +306,34 @@ func (c *bigQueryClient) ListAccess(ctx context.Context, resources []*domain.Res
 	return access, nil
 }
 
+func (c *bigQueryClient) getGrantAbleRoleForDataset() ([]bq.AccessRole, error) {
```
mabdh, line 309: does it mean the `grantable role` ?

plugins/providers/bigquery/client.go
```diff
@@ -327,6 +306,34 @@ func (c *bigQueryClient) ListAccess(ctx context.Context, resources []*domain.Res
...
+	var roles = []bq.AccessRole{bq.OwnerRole, bq.WriterRole, bq.ReaderRole}
+	var resourceName string
+	ctx := context.Background()
+	datasetIterator := c.client.Datasets(ctx)
+	for {
+		dataset, err := datasetIterator.Next()
+		if err == iterator.Done || err != nil {
```
mabdh, line 316: This one `iterator` or `datasetIterator` ?

plugins/providers/bigquery/client.go
```diff
@@ -327,6 +306,34 @@ func (c *bigQueryClient) ListAccess(ctx context.Context, resources []*domain.Res
...
+
+		for _, role := range response.Roles {
+			roles = append(roles, bq.AccessRole(role.Name))
+		}
+		return roles, nil
+	}
+	return roles, nil
+}
+
 func (c *bigQueryClient) getGrantableRolesForTables() ([]string, error) {
```
mabdh, line 337: Why this return `[]string` while `getGrantableRolesForDataset` return []bq.AccessRole` ?

plugins/providers/bigquery/config.go
```diff
@@ -173,7 +169,18 @@ func (c *Config) validatePermission(value interface{}, resourceType string, clie
...
+		if err != nil {
+			return nil, ErrCannotFetchDatabasePermission
+		}
+		findMatchingRole := false
+		for _, role := range roles {
+			if string(role) == permision {
+				findMatchingRole = true
+				break
+			}
+		}
```
mabdh, line 182: can we use `strings.Contains` for this?

plugins/providers/bigquery/client.go
```diff
@@ -327,6 +306,34 @@ func (c *bigQueryClient) ListAccess(ctx context.Context, resources []*domain.Res
 	return access, nil
 }
 
+func (c *bigQueryClient) getGrantAbleRoleForDataset() ([]bq.AccessRole, error) {
+	var roles = []bq.AccessRole{bq.OwnerRole, bq.WriterRole, bq.ReaderRole}
+	var resourceName string
+	ctx := context.Background()
```
mabdh, line 312: why create a new `context` instead of propagating it?

plugins/providers/bigquery/client.go
singhvikash11: this one datasetIterator

plugins/providers/bigquery/config.go
singhvikash11: any specific reason to use contains

mabdh: Please correct me if I am wrong, this one is only try to find whether the `permision` inside `roles` right?

plugins/providers/bigquery/client.go
```diff
@@ -110,38 +110,21 @@ func (c *bigQueryClient) GetTables(ctx context.Context, datasetID string) ([]*Ta
...
 	for _, a := range metadata.Access {
-		if a.Entity == user && a.Role == bqRole {
+		if a.Entity == user && a.Role == bq.AccessRole(role) {
 			return ErrPermissionAlreadyExists
 		}
 	}
 	update := bq.DatasetMetadataToUpdate{
 		Access: append(metadata.Access, &bq.AccessEntry{
-			Role:       bqRole,
+			Role:       bq.AccessRole(role),
```
rahmatrhd, line 127: please make sure granting a role such as `roles/bigquery.dataViewer` (one example returned from queryGrantableRoles) instead of `READER` is possible using this API, because as per the client library codes, `bq.AccessRole` are only READER, WRITER, and OWNER

plugins/providers/bigquery/client.go
```diff
@@ -110,38 +110,21 @@ func (c *bigQueryClient) GetTables(ctx context.Context, datasetID string) ([]*Ta
...
 		return err
 	}
 
-	bqRole, err := c.ResolveDatasetRole(role)
-	if err != nil {
-		return err
-	}
 	for _, a := range metadata.Access {
-		if a.Entity == user && a.Role == bqRole {
+		if a.Entity == user && a.Role == bq.AccessRole(role) {
```
rahmatrhd, line 121: is it possible if `a.Role="READER"` and `role="roles/bigquery.dataViewer"`? if so, it should still fulfill the condition checking here in a way because both are essentially the same role

plugins/providers/bigquery/client.go
singhvikash11: @mabdh there is no `context` from the caller of this method

`(p *Provider) CreateConfig(pc *domain.ProviderConfig) error`

To avoid change the signature of it creating a context in method itself

plugins/providers/bigquery/client.go
singhvikash11: @rahmatrhd here we are adding all possible roles on BigQuery and allowed roles on datasets READER, WRITER, and OWNER
`var roles = []bq.AccessRole{bq.OwnerRole, bq.WriterRole, bq.ReaderRole}`

here we are adding these default roles to roles to not missed these roles

plugins/providers/bigquery/client.go
singhvikash11: @rahmatrhd it'll not be the same, users should not use both to avoid ambiguity.

plugins/providers/bigquery/config.go
singhvikash11: @mabdh no it's not, it's checking for exact match

plugins/providers/bigquery/client.go
singhvikash11: bq.AccessRole is a go type for string and I don't want to touch tables flow in this pr.

plugins/providers/bigquery/client.go
singhvikash11: done

plugins/providers/bigquery/client.go
singhvikash11: handle the scenario

plugins/providers/bigquery/client.go
singhvikash11: handle the scenario

plugins/providers/bigquery/client.go
```diff
@@ -327,6 +306,34 @@ func (c *bigQueryClient) ListAccess(ctx context.Context, resources []*domain.Res
 	return access, nil
 }
 
+func (c *bigQueryClient) getGrantableRolesForDataset() ([]bq.AccessRole, error) {
+	var roles = []bq.AccessRole{bq.OwnerRole, bq.WriterRole, bq.ReaderRole}
+	var resourceName string
+	ctx := context.Background()
+	datasetIterator := c.client.Datasets(ctx)
+	for {
```
rahmatrhd, line 314: I think `for` loop here is not necessary?

plugins/providers/bigquery/client.go
rahmatrhd: should this be fixed?

plugins/providers/bigquery/client.go
rahmatrhd, line 127: can you check if granting access with role such as `bq.AccessRole("roles/bigquery.metadataViewer")` (or anything else that is not equivalent to VIEWER/WRITER/OWNER) would work using this `client`? Otherwise we might need to use the `apiClient` for granting/revoking dataset

rahmatrhd (changes requested):

plugins/providers/bigquery/client.go
singhvikash11: fixed

plugins/providers/bigquery/client.go
singhvikash11: fixed

plugins/providers/bigquery/provider.go
```diff
@@ -219,6 +217,24 @@ func (p *Provider) GetRoles(pc *domain.ProviderConfig, resourceType string) ([]*
 	return provider.GetRoles(pc, resourceType)
 }
 
+func (p *Provider) GetPermissions(pc *domain.ProviderConfig, resourceType, role string) ([]interface{}, error) {
```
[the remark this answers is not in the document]
singhvikash11: @rahmatrhd bigquery provider has a custom method to resolve the same permissions like `roles/bigquery.dataViewer and READER`

[custom method](https://github.com/odpf/guardian/blob/e6aed37dc5468af0862a3a038fb6a15ec0eb37e2/plugins/providers/bigquery/model.go#L86)

plugins/providers/bigquery/client.go
singhvikash11: fixed

plugins/providers/bigquery/client.go
singhvikash11: granting access like `roles/bigquery.metadataViewer` is working with existing client

singhvikash11:
@rahmatrhd let's close this pr, I guess you already fixed it by 
getGcloudPermissions
translateDatasetRoleToBigQueryRole

Issue #311, opened 2022-10-19
refactor appeal functions


rahmatrhd:
proposing to refactor these items to improve code readability esp. in appeal and approval services

### create appeal
- https://github.com/odpf/guardian/blob/075f577dd0d1583a165917121f4517411bc8cf40/core/appeal/service.go#L909 move `appeal.Service.fillApprovals` to `domain.Appeal.InitApprovals(p domain.Policy) error`
- add `domain.Step.ToApproval(approvers []string) domain.Approval` to automatically convert list of approval step config to `domain.Approval`
- https://github.com/odpf/guardian/blob/075f577dd0d1583a165917121f4517411bc8cf40/core/appeal/service.go#L768 move `appeal.Service.resolveApprovers` to `domain.Step.ResolveApprovers(a domain.Appeal)`
- https://github.com/odpf/guardian/blob/075f577dd0d1583a165917121f4517411bc8cf40/core/appeal/service.go#L976 rename `appeal.Service.CreateAccess` to `appeal.Service.GrantAccessToProvider`
- add unit tests for methods in `domain.Appeal` and `domain.Approval`

### make action/update approval
- https://github.com/odpf/guardian/blob/075f577dd0d1583a165917121f4517411bc8cf40/core/appeal/service.go#L376 move `appeal.Service.MakeAction` to `approval.Service.UpdateApproval`

### approval. advance approval
- https://github.com/odpf/guardian/blob/075f577dd0d1583a165917121f4517411bc8cf40/core/approval/service.go#L50 move `approval.Service.AdvanceApproval` to `domain.Appeal.AdvanceApproval(p domain.Policy)`

mabdh:
It looks like the changes are not really a major changes. Regardless I think it is a good start to refactor the codebase into odpf-like folder structure.

rahmatrhd:
@mabdh what do you mean by odpf-like folder structure?

mabdh:
@rahmatrhd what I meant was folder-structure of odpf standard 
similar with #199

PR #322, opened 2022-11-08
refactor: appeal service

haveiss:
Closes https://github.com/odpf/guardian/issues/311


core/appeal/service.go
```diff
@@ -254,14 +249,14 @@ func (s *Service) Create(ctx context.Context, appeals []*domain.Appeal, opts ...
...
 		}
 
-		if err := s.fillApprovals(appeal, policy); err != nil {
+		if err := appeal.InitApprovals(policy); err != nil {
 			return fmt.Errorf("populating approvals: %w", err)
 		}
 
 		appeal.Init(policy)
 
 		appeal.Policy = policy
```
mabdh, line 258: Can these 3 be merged together in this flow? @haveiss @rahmatrhd 
I assume `appeal.Init` should init all.

core/approval/service.go
```diff
@@ -47,110 +123,284 @@ func (s *Service) BulkInsert(ctx context.Context, approvals []*domain.Approval)
...
+		Email:      email,
+	})
+}
+
+func (s *Service) DeleteApprover(ctx context.Context, approvalID, email string) error {
+	return s.repo.DeleteApprover(ctx, approvalID, email)
+}
+
+// UpdateApproval updates an approval.
+func (s *Service) UpdateApproval(ctx context.Context, approvalAction domain.ApprovalAction) (*domain.Appeal, error) {
```
rahmatrhd, line 138: seeing the dependecies to other services makes me think update approval still should be part of appealService 🤔

rahmatrhd (changes requested):

core/appeal/service.go
rahmatrhd: https://github.com/odpf/guardian/pull/322#discussion_r1024852918

we can group this into an appeal method called `ApplyPolicy` or something similar. Because these logic mostly about initializing the appeal based on the policy

domain/appeal.go
```diff
@@ -142,6 +153,181 @@ func (a Appeal) ToGrant() (*Grant, error) {
...
+			approverEmails, err = a.ResolveApprovers(step.Approvers)
+			if err != nil {
+				return fmt.Errorf("resolving approvers `%s`: %w", step.Approvers, err)
+			}
+		}
+
+		approval := &Approval{}
+		if err := approval.Init(p, i, approverEmails); err != nil {
+			return fmt.Errorf(`initializing approval "%s": %w`, step.Name, err)
+		}
```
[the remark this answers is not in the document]
haveiss: @mabdh do you have any suggestion on how to implement the `chain of responsibility` pattern here?

core/approval/service.go
haveiss: yea, let met thing of something, or do you have any suggestion to make this better?

haveiss: moving back update approval to appealService to reduce dependencies

domain/appeal.go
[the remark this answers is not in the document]
haveiss: refactored using suggestion 2 from mas @rahmatrhd

core/appeal/service.go
```diff
@@ -391,125 +385,135 @@ func (s *Service) MakeAction(ctx context.Context, approvalAction domain.Approval
...
-			if approval.Status != domain.ApprovalStatusPending {
-				if err := checkApprovalStatus(approval.Status); err != nil {
-					return nil, err
-				}
-			}
+		}
 
-			if !utils.ContainsString(approval.Approvers, approvalAction.Actor) {
-				return nil, ErrActionForbidden
+		if approval.Status != domain.ApprovalStatusPending {
```
rahmatrhd, line 390: moving this code out from the `else` block will change the logic as well. Have we tested this?

core/appeal/service.go
```diff
@@ -972,8 +888,7 @@ func (s *Service) handleAppealRequirements(ctx context.Context, a *domain.Appeal
 	return nil
 }
 
-func (s *Service) CreateAccess(ctx context.Context, a *domain.Appeal, opts ...CreateAppealOption) error {
-	// TODO: rename to GrantAccess
+func (s *Service) GrantAccessToProvider(ctx context.Context, a *domain.Appeal) error {
```
rahmatrhd, line 891: why do we remove the options from this func parameter?

haveiss: previously removing because the `opts` arguments never passed anywhere on this [commit](https://github.com/odpf/guardian/commit/015bec7020d4613e61f26d7e0e0879d683dbd681), seems unused and make the mocking the function become complex,

should we add this back?

core/appeal/service.go
haveiss: yes, existing tests are working with high coverages, inside the `if` there only 2 possibilities, either `return` or `continue`, it won't reach next lines, 
so we can remove the else here

https://github.com/odpf/guardian/blob/ae6b057ad98d1efefc00d8b0957747670efacfe7/core/appeal/service.go#L392-L397

core/appeal/service.go
rahmatrhd: It was added [here](https://github.com/odpf/guardian/pull/222/files#diff-4ceaf11357ea8ac162487cfbb8f9f0a0a7446fa7b61bc799aac1a3d3ca7adc58R264) and got accidentally removed [here](https://github.com/odpf/guardian/pull/249/files#diff-4ceaf11357ea8ac162487cfbb8f9f0a0a7446fa7b61bc799aac1a3d3ca7adc58R282) I think. @bsushmith could you help here?

haveiss: nice catch mas @rahmathd, added the opts back

rahmatrhd (approved):

PR #323, opened 2022-11-08
feat: add resources conditionally

### LINKED ISSUE #313: Add resources conditionally
this issue is converted from this [RFC PR](https://github.com/odpf/guardian/pull/156) 

# Summary

Resource in Guardian represents the actual resource in the provider e.g. for BigQuery provider, a resource represents a
dataset or a table. Guardian collects resources from the provider automatically as soon as it registered. While in
parallel, Guardian also has a job for continuously syncing resources.

In current design it fetch all the database/table/resources in configure provider, there is no way to filter out on
resources.

# Technical Design

Add a scope of condition in provider configuration to filter out on resources.

#### Using expression

Expression can be used to check whether Guardian should add the resources or not.

```yaml
resources:
  filter: $resource.urn endsWith 'keyset'
  type: tables
  policy:
    id: keyset_table_policy
```

#### Migration

We can edit or add new providers based on new configuration. For already added resources whom we want to make conditional need to soft-delete it from Guardian db.

singhvikash11:
Closes  #313


domain/provider.go
```diff
@@ -40,6 +40,7 @@ type PolicyConfig struct {
 // ResourceConfig is the configuration for a resource type within a provider
 type ResourceConfig struct {
 	Type   string        `json:"type" yaml:"type" validate:"required"`
+	Filter string        `json:"filter" yaml:"type"`
```
bsushmith, line 43: yaml tag is incorrect.

docs/docs/providers/bigquery.md
```diff
@@ -71,6 +71,7 @@ resources:
         permissions:
           - WRITER
   - type: table
+    filter: $resource.name endsWith transtaction
```
bsushmith, line 74: Can you fix the typo in here - `transaction`

core/provider/service_test.go
```diff
@@ -312,6 +312,43 @@ func (s *ServiceTestSuite) TestFetchResources() {
...
 	})
+
+	s.Run("should upsert filter resources on success", func() {
+		providersWithResourceFilter := []*domain.Provider{
+			{
+				ID:   "1",
+				Type: mockProviderType,
+				URN:  mockProvider,
+				Config: &domain.ProviderConfig{Resources: []*domain.ResourceConfig{
+					{Type: "dataset", Filter: "$resource.urn == 'resource2'"},
```
bsushmith, line 323: Can you add one more test case with `$resource.name endsWith transaction`  as the filter criteria

core/provider/service.go
```diff
@@ -564,3 +595,20 @@ func validateDuration(d string) error {
 	}
 	return nil
 }
+
+func structToMap(item interface{}) (map[string]interface{}, error) {
```
bsushmith, line 599: Can you move this generic function to utils?

core/provider/service.go
```diff
@@ -450,14 +453,42 @@ func (s *Service) getResources(ctx context.Context, p *domain.Provider) ([]*doma
...
+				return nil, fmt.Errorf("parsing resource struct to map: %w", err)
+			}
+
+			v, err := evaluator.Expression(filterExpression).EvaluateWithVars(map[string]interface{}{
+				"resource": resourceMap,
+			})
+			if err != nil {
+				return nil, err
+			}
+			if !reflect.ValueOf(v).IsZero() {
```
bsushmith, line 481: I'm not sure if this also checks if `v` is nil. Can you please check and add a `IsNil` check for `v`?

singhvikash11: ValueOf check for nil

bsushmith (approved):

rahmatrhd (changes requested):
let's add a validation on provider create&update for this new `filter` field. We can create a dummy resource and try to evaluate the expression onto that resource to see if the expression if valid and working or not

pkg/evaluator/expression.go
```diff
@@ -23,6 +24,23 @@ func (m ExprParam) Split(s, sep string) []string {
 	return strings.Split(s, sep)
 }
 
+func (e Expression) EvaluateWithStruct(item interface{}) (interface{}, error) {
```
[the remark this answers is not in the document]
singhvikash11: @mabdh @rahmatrhd test is already there for this
https://github.com/odpf/guardian/blob/filter_resource/core/provider/service_test.go#L323

singhvikash11:
> let's add a validation on provider create&update for this new `filter` field. We can create a dummy resource and try to evaluate the expression onto that resource to see if the expression if valid and working or not

https://github.com/odpf/guardian/blob/filter_resource/core/provider/service_test.go#L323
I have verified this manually also

singhvikash11:
@rahmatrhd @mabdh @bsushmith please review changes and approve the pr so we can merge it

ravisuhag:
@rahmatrhd Please see if we can review and close this.

rahmatrhd (approved):
LGTM

Issue #299, opened 2022-09-29
Appeal revoke on deleted resources is not working


singhvikash11:
**Description**
failed to revoke an appeal for an underlying resource if the resource got renamed or deleted. We have access report comparison where we have requested in guardian doesn't exist into GCP big query. These resources don't exist in guardian since people delete the dataset or renamed it and revoke of appeal failed with an error to find the same resource in GCP. 

**Expected behaviour**
Either soft-delete on a resource during fetch resources will revoke these appeals or the appeal can handle in case a resource doesn't exist.

bsushmith:
@singhvikash11 can a resource be deleted and recreated later with the same name? 

For example, in `bigquery`, if one deletes a dataset named - `city_flags`, and later someone else creates a dataset with same name... 

In this scenario, I would assume - that guardian would revoke appeals at the time when the initial dataset was deleted.
Because if it did not revoke at the time of resource deletion, the dataset recreated later in `bigquery` would not have the user access to it but the guardian `grants` table would show that the user still has access.


@rahmatrhd @mabdh WDYT?

PR #327, opened 2022-11-16
fix: fix grant revoke for deleted resource into guardian.

singhvikash11:
fix: #299

### DIFF
--- a/internal/store/postgres/migrations/000012_update_status_provider_for_appeal_source.up.sql
+++ b/internal/store/postgres/migrations/000012_update_status_provider_for_appeal_source.up.sql
@@ -1,0 +1,7 @@
+UPDATE
+    "grants"
+SET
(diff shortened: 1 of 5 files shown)


internal/store/postgres/migrations/000012_update_status_provider_for_appeal_source.up.sql
```diff
@@ -0,0 +1,7 @@
+UPDATE
+    "grants"
+SET
+    "status_in_provider" = "status"
+WHERE
+    "source" = 'appeal'
+    AND "status_in_provider" != "status"
```
rahmatrhd, line 7: if I assume it correctly this is for fixing the previously wrong-assigned data(?) If so,  can we not do it through app migration but run this SQL fix in our running db instances directly?

PR #328, opened 2022-11-17
fix: make notification as local variable

### LINKED ISSUE #329: Getting 3 times slack notification for same resource reminder
**Description**
For the same resource expiration guardian send 3 times notification, I am guessing the reordering of var notifications can cause issues.

image

singhvikash11:
fix #329

### DIFF
--- a/jobs/grant_expiration_reminder.go
+++ b/jobs/grant_expiration_reminder.go
@@ -34,7 +34,7 @@ func (h *handler) GrantExpirationReminder(ctx context.Context) error {
 		}
 
 		// TODO: group notifications by username
-		var notifications []domain.Notification
+		notifications := make([]domain.Notification, 0)
 		for _, g := range grants {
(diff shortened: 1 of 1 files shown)


mabdh:
Have we tested this? Or could we test this flow to make sure this is a correct solution?

mabdh:
This should be a `fix` instead a `chore` right?

rahmatrhd:
can you also share how this changes could fix the issue? wanted to know why the previous code causing the issue 🙏

bsushmith:
This PR is no longer needed.

ravisuhag:
@bsushmith Is the issue #329 solved in another PR or that still remains open? cc @rahmatrhd

bsushmith:
@ravisuhag It is still open - We haven't aligned on when we want to deprecate the older configs (which were present for backward compatibility)
comment - https://github.com/odpf/guardian/issues/329#issuecomment-1321448049

PR #330, opened 2022-11-18
feat: Initial implementation for slack Api

singhvikash11:


ravisuhag:
@singhvikash11 Slack for approvals etc. should act like a client only. We should not club it with the server code. Can you first describe and discuss your approach on an issue. So that we are aligned on approach.

PR #331, opened 2022-11-22
feat: introduce provider activity 

rahmatrhd:


Makefile
```diff
@@ -3,7 +3,7 @@ COMMIT := $(shell git rev-parse --short HEAD)
 TAG := "$(shell git rev-list --tags --max-count=1)"
 VERSION := "$(shell git describe --tags ${TAG})-next"
 BUILD_DIR=dist
-PROTON_COMMIT := "6e5d6390c14e13815269e536ec5d967eb5202421"
+PROTON_COMMIT := "f8df60d95566d662e2fae96fee483dd17b487b0b"
```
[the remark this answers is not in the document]
rahmatrhd: created a new proton PR, kindly check as well https://github.com/odpf/proton/pull/232

plugins/providers/bigquery/provider.go
```diff
@@ -255,6 +340,98 @@ func (p *Provider) getBigQueryClient(credentials Credentials) (BigQueryClient, e
...
+	case DatasetRoleReader:
+		roleID = "roles/bigquery.dataViewer"
+	}
+
+	if permissions, exists := p.rolesCache.Get(roleID); exists {
+		p.logger.Debug("getting permissions from cache", "role", roleID)
+		return permissions.([]string), nil
+	}
+
+	p.logger.Debug("getting permissions from gcloud", "role", roleID)
```
haveiss, line 415: how about change from debug to info?

plugins/providers/bigquery/provider.go
```diff
@@ -255,6 +340,98 @@ func (p *Provider) getBigQueryClient(credentials Credentials) (BigQueryClient, e
...
+func (p *Provider) getGcloudPermissions(ctx context.Context, pd domain.Provider, gcloudRole string) ([]string, error) {
+	roleID := gcloudRole
+	switch gcloudRole {
+	case DatasetRoleOwner:
+		roleID = "roles/bigquery.admin"
+	case DatasetRoleWriter:
+		roleID = "roles/bigquery.dataEditor"
+	case DatasetRoleReader:
+		roleID = "roles/bigquery.dataViewer"
+	}
```
haveiss, line 408: let's extract this to new function

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,99 @@
+package cache
```
haveiss, line 1: noob question, why we implement the cache ourself instead of using third party?

plugins/providers/bigquery/client.go
```diff
@@ -327,6 +327,26 @@ func (c *bigQueryClient) ListAccess(ctx context.Context, resources []*domain.Res
...
+
+	if strings.HasPrefix(role, "roles/") {
+		iamRole, err = c.iamService.Roles.Get(role).Context(ctx).Do()
+	} else if strings.HasPrefix(role, "projects/") {
+		iamRole, err = c.iamService.Projects.Roles.Get(role).Context(ctx).Do()
+	} else if strings.HasPrefix(role, "organizations/") {
+		iamRole, err = c.iamService.Organizations.Roles.Get(role).Context(ctx).Do()
+	} else {
+		return nil, fmt.Errorf("invalid role signature: %q", role)
+	}
```
haveiss, line 342: [nit] how about to use switch for better readability?

internal/store/postgres/activity_repository.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+		}
+
+		return nil
+	})
+}
+
+func getResourceUniqueURN(r model.Resource) string {
+	return fmt.Sprintf(`%s/%s/%s/%s`, r.ProviderType, r.ProviderURN, r.Type, r.URN)
+}
```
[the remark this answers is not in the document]
rahmatrhd: was also thinking the same way, but didn't go with that because this helper is only being used within this Upsert method, having it as the model's method will make it available to other places which using the model and was trying to avoid that only

pkg/cache/inmemory.go
rahmatrhd: only because it's a simple in memory cache, do you have any recommended package for in mem cache?

haveiss: can find in [here](https://github.com/avelino/awesome-go#caches), but this one is quite popular
- https://github.com/patrickmn/go-cache
- https://github.com/allegro/bigcache

haveiss (commented):
can you rebase? the PR had conflict

haveiss (approved):

mabdh (approved):

PR #338, opened 2022-12-09
feat: migrate to salt mux

haveiss:
closes https://github.com/odpf/guardian/issues/68

### DIFF
--- a/internal/server/services.go
+++ b/internal/server/services.go
@@ -56,7 +56,10 @@ func InitServices(deps ServiceDeps) (*Services, error) {
 		return nil, err
 	}
 
-	sqldb, _ := store.DB().DB()
+	sqldb, err := store.DB().DB()
+	if err != nil {
+		return nil, err
+	}
 
 	auditRepository := audit_repos.NewPostgresRepository(sqldb)
 	auditLogger := audit.New(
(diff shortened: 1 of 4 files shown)


internal/server/services.go
```diff
@@ -54,7 +54,9 @@ func InitServices(deps ServiceDeps) (*Services, error) {
 		return nil, err
 	}
 
-	auditRepository := audit_repos.NewPostgresRepository(store.DB())
+	sqldb, _ := store.DB().DB()
+
+	auditRepository := audit_repos.NewPostgresRepository(sqldb)
```
rahmatrhd, line 59: we also need to make sure the audit's migration is being called from our migration

internal/server/services.go
```diff
@@ -54,7 +54,9 @@ func InitServices(deps ServiceDeps) (*Services, error) {
 		return nil, err
 	}
 
-	auditRepository := audit_repos.NewPostgresRepository(store.DB())
+	sqldb, _ := store.DB().DB()
```
rahmatrhd, line 57: let's catch and return the `err`

rahmatrhd (changes requested):

rahmatrhd (approved):

PR #341, opened 2023-01-09
feat(opentelemetry): enable tracing

haveiss:


mabdh (approved):

bsushmith:
@haveiss @mabdh Can we have an issue for this with details on the scope of this change?
Would help in understanding the changes better.

Issue #345, opened 2023-01-13
bigquery client throwing error `CREDENTIALS_MISSING`


bsushmith:
**Description**
Bigquery client is throwing error for all bigquery api's in the latest guardian code. This is due to the telemetry changes that have been done at the bigquery client initialization level. 

this is how the client is being initialized in code... we are passing the telemetry http client along with the credentials option.
```go
func newBigQueryClient(projectID string, credentialsJSON []byte) (*bigQueryClient, error) {
	ctx := context.Background()
	clientOpts := []option.ClientOption{
		option.WithCredentialsJSON(credentialsJSON),
		option.WithHTTPClient(tracing.NewHttpClient("BigQueryHttpClient")),
	}
	client, err := bq.NewClient(ctx, projectID, clientOpts...)
	if err != nil {
		return nil, err
	}
```
but based on [docs](https://pkg.go.dev/google.golang.org/api/option#WithHTTPClient), it states that the httpclient option overrides all other supplied options. this httpclientoption is overriding the credentials clientoption as bqClient is assuming client is already supplied through this second telemetry option 

```go
// WithHTTPClient returns a ClientOption that specifies the HTTP client to use
// as the basis of communications. This option may only be used with services
// that support HTTP as their communication transport. When used, the
// WithHTTPClient option takes precedent over all other supplied options.
func WithHTTPClient(client *http.Client) ClientOption {
	return withHTTPClient{client}
}
```

**To Reproduce**
* run server on latest main
* hit any bigquery plugin related api

**Additional context**
branch - `main`

@utsav-agarwal

bsushmith:
We've fixed it by reverting the code GCP related providers. I suggest we keep this issue open till we address the main issue and the changes again in a different way

haveiss:
> We've fixed it by reverting the code GCP related providers. I suggest we keep this issue open till we address the main issue and add the changes again in a different way 
> 
> @haveiss

thanks! we can close this issue

PR #346, opened 2023-01-13
fix: reverted tracing ccode in client code for bigquery and gcloud

utsav14nov:

PR #347, opened 2023-01-13
fix: gcloud client tracing

haveiss:
fix https://github.com/odpf/guardian/issues/345

If we use WithHTTPClient(), then the BigQuery client will be created without applying the WithToken() option. This means that Oauth won't be set up properly.


haveiss:
duplicates with https://github.com/odpf/guardian/pull/346

PR #349, opened 2023-01-17
feat(idtoken-auth): add auth validator using google idtoken

swallowstalker:
Add google idtoken (OIDC) auth validator on grpc interceptor, to provide login via google SSO

# How to use it

Just add this line to config.yaml (sample is on internal/server/config.yaml)

```yaml
AUTH:
  PROVIDER: default   # can be "default" or "oidc"
  DEFAULT:
    # AUTHENTICATED_USER_HEADER_KEY takes priority for backward-compatibility
    HEADER_KEY: X-Auth-Email   
  OIDC:
    # add google client audience
    AUDIENCE: "some-kind-of-audience.com"  
    # add eligible email domains for your organization, empty string means that no checking will be done
    ELIGIBLE_EMAIL_DOMAINS: "emaildomain1.com,emaildomain2.com"
```

# Tests

## Without any request header

```
curl --location --request GET 'http://localhost:3000/api/v1beta1/appeals'
```

Response: `401 Unauthorized`
```
{
    "code": 16,
    "message": "invalid authentication credentials"
}
```

## With `Authorization` header containing `idtoken` from google SSO

```
curl --location --request GET 'http://localhost:3000/api/v1beta1/appeals' \
--header 'Authorization: Bearer <idtoken>'
```

Response: `200 OK` (no list of appeals yet because it's empty)
```
{}
```


bsushmith:
@swallowstalker can you create an issue along with details what this feature is about and what it will solve, and link it with this PR ?

swallowstalker:
> @swallowstalker can you create an issue along with details what this feature is about and what it will solve, and link it with this PR ?

sure, will create it shortly. sorry if there's only very short description and background about this PR

swallowstalker:
@bsushmith here is the issue https://github.com/odpf/guardian/issues/350

internal/server/config.go
```diff
@@ -56,5 +72,11 @@ func LoadConfig(configFile string) (Config, error) {
 		}
 		return Config{}, err
 	}
+
+	// keep for backward-compatibility
+	if cfg.AuthenticatedUserHeaderKey != "X-Auth-Email" {
```
mabdh, line 77: is this really necessary to manually check this? if we have defined default tag, the value will be used if we let this empty.
I think it is even not necessary to define the default value tag.

pkg/auth/oidc.go
```diff
@@ -0,0 +1,109 @@
...
+	"google.golang.org/grpc/status"
+)
+
+var InvalidAuthError = status.Errorf(codes.Unauthenticated, "invalid authentication credentials")
+
+type Validator interface {
+	Validate(ctx context.Context, token string, audience string) (*idtoken.Payload, error)
+}
+
+type OidcValidator struct {
```
mabdh, line 22: Calling this as `OIDCValidator` for acronym would be more suitable according to [this](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

pkg/auth/oidc.go
```diff
@@ -0,0 +1,109 @@
...
+
+type OidcValidatorParams struct {
+	Audience          string
+	ValidEmailDomains string
+	HeaderKey         string
+	ContextKey        interface{}
+}
+
+func NewOidcValidator(validator Validator, config *OidcValidatorParams) *OidcValidator {
+	audience := config.Audience
```
mabdh, line 38: this will throw panic if config is nil. is it necessary to pass config as pointer?

swallowstalker: nice catch, I have modified it into non-pointer one to force user to call it by non-nil params

pkg/auth/oidc.go
swallowstalker: that's right, i've modified it, thanks!

internal/server/config.go
swallowstalker: the idea was to check `cfg.AuthenticatedUserHeaderKey` if it contains non-default value (such as user configure the header key on older version, let's say `X-Something`), then `cfg.Auth.Default.HeaderKey` will automatically be assigned the old value (to avoid `X-Something` accidentally changed to `X-Auth-Email`)

however, after giving it some thought, i think that's a good idea to let `cfg.AuthenticatedUserHeaderKey` without default tag so to compare it with empty string. do you have suggestions on make this backward-compatibility support better?

ravisuhag:
@rahmatrhd @haveiss @mabdh can we check this and merge if it looks good?

ravisuhag:
@swallowstalker one more thing i noticed, we can move the mocks in the same package for which it is used. You can refer to other mocks implementations and placement.

internal/server/config.go
```diff
@@ -33,6 +33,21 @@ type Jobs struct {
 	ExpiringAccessNotification JobConfig `mapstructure:"expiring_access_notification"`
 }
 
+type DefaultAuth struct {
+	HeaderKey string `mapstructure:"header_key" default:"X-Auth-Email"`
+}
+
+type OIDCAuth struct {
```
ravisuhag, line 40: @swallowstalker would it make sense to move this config next to OIDC auth only? and server config can import it from there.

swallowstalker:
> @swallowstalker one more thing i noticed, we can move the mocks in the same package for which it is used. You can refer to other mocks implementations and placement.

I have moved OIDCValidator mock to `pkg/auth/mocks` directory 👍

internal/server/config.go
swallowstalker: hi @ravisuhag, I've moved the OIDCAuth config to pkg/auth, i think it's more clear that way

internal/server/server.go
```diff
@@ -266,3 +274,27 @@ func fetchDefaultJobScheduleMapping() map[JobType]string {
...
+	if config.Auth.Provider == "oidc" {
+		idtokenValidator, err := idtoken.NewValidator(context.Background())
+		if err != nil {
+			return nil, err
+		}
+
+		params := auth.OIDCValidatorParams{
+			Audience:          config.Auth.OIDC.Audience,
+			ValidEmailDomains: config.Auth.OIDC.EligibleEmailDomains,
+			HeaderKey:         config.Auth.Default.HeaderKey,
```
rahmatrhd, line 291: can OIDC config has its own `HeaderKey` field so we don't have to add `auth.default.*` when using `auth.provider = oidc`?

rahmatrhd (changes requested):

internal/server/config.go
```diff
@@ -33,6 +34,16 @@ type Jobs struct {
...
+	HeaderKey string `mapstructure:"header_key" default:"X-Auth-Email"`
+}
+
+type Auth struct {
+	Provider string        `mapstructure:"provider" default:"default"`
+	Default  DefaultAuth   `mapstructure:"default"`
+	OIDC     auth.OIDCAuth `mapstructure:"oidc"`
+}
+
 type Config struct {
```
rahmatrhd, line 47: let's also add `// Deprecated: use Auth.Default.HeaderKey instead` note on the `AuthenticatedUserHeaderKey` to mark the field as deprecated and suggest to use the new one

internal/server/server.go
ravisuhag: That make sense. each auth provider should have their own `HeaderKey`.

ravisuhag:
@swallowstalker Please see if you can find some time to fix test and the above mentioned comment. We can go ahead and merge it then

swallowstalker:
> @swallowstalker Please see if you can find some time to fix test and the above mentioned comment. We can go ahead and merge it then

sorry i was oncall last week. will fix that in few days

internal/server/server.go
swallowstalker: hi @rahmatrhd @ravisuhag, I've refactored header key and (additionally) context key. let me know if you have any concerns

pkg/auth/oidc.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/sirupsen/logrus"
+	"google.golang.org/api/idtoken"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+var InvalidAuthError = status.Errorf(codes.Unauthenticated, "invalid authentication credentials")
+var OIDCHeaderKey = "X-OIDC-Email"
```
rahmatrhd, line 17: let's make it configurable via `OIDCAuth.HeaderKey` instead of hardcoding it

internal/server/server.go
```diff
@@ -266,3 +274,25 @@ func fetchDefaultJobScheduleMapping() map[JobType]string {
...
+		if err != nil {
+			return nil, err
+		}
+
+		params := auth.OIDCValidatorParams{
+			Audience:          config.Auth.OIDC.Audience,
+			ValidEmailDomains: config.Auth.OIDC.EligibleEmailDomains,
+		}
+
+		bearerTokenValidator := auth.NewOIDCValidator(idtokenValidator, params)
```
rahmatrhd, line 293:
```suggestion
		bearerTokenValidator := auth.NewOIDCValidator(idtokenValidator, config.Auth.OIDC)
```

is it possible if we just use oidc config and get rid of `OIDCValidatorParams`?

rahmatrhd (changes requested):

pkg/auth/oidc.go
swallowstalker: after giving it some thought, I only saw two usage of header key (containing email). First at `ctx_logrus.AddFields` [code](https://github.com/odpf/guardian/blob/main/internal/server/auth.go#L21). I think it's used for audit log. 

second, it is located at `getUser` [code](https://github.com/odpf/guardian/blob/main/api/handler/v1beta1/grpc.go#L154). I realized it because when rejecting an appeal, it still needs email on header request (if I choose OIDC auth)

as for first location, it is already solved by getting email from OIDC JWT payload, then set logrus fields key to `OIDCHeaderKey`. I think we can standardize this logrus field somehow to a default value (same with default auth logrus field)

as for second location, I propose that we set context value to `getUser` with two values (it will be set on interceptor)
- auth type
- context key
then we can get the user from context instead of getting it from metadata header request. How does that sound?

rahmatrhd: interesting, for second location, we can also just use a single context value containing the authentication email (like the one in first location). And the interceptor will set the value according to the auth type accordingly. 

With this approach, we can also set the logrus field only once since the prior auth interceptor has set the authentication email context value like I mentioned above. What do you think?

swallowstalker: i think that's good. i will modify it shortly

swallowstalker: hi @rahmatrhd i have adjust the code, do you have any comments on that?

internal/server/server.go
swallowstalker: that's right. i forgot why i used that params but i have refactored it anyway 😂

rahmatrhd (approved):
LGTM

rahmatrhd:
@swallowstalker I've created [v0.6.0-alpha.1](https://github.com/odpf/guardian/releases/tag/v0.6.0-alpha.1) release, you can use this to test from your side 🙂

ravisuhag:
@swallowstalker Thanks a lot for your contribution :)

PR #353, opened 2023-01-27
feat: bump salt config version

haveiss:
### Context
- Relates with https://github.com/odpf/salt/pull/66
- To fix telemetry headers configuration via environment variable


rahmatrhd (approved):
should we wait until the salt PR is merged into main and then bump to that version?

haveiss:
> should we wait until the salt PR is merged into main and then bump to that version?

latest salt version are breaking changes https://github.com/odpf/salt/pull/63, it requires to make changes, let's do that on separate PR, wdyt @rahmatrhd ?

PR #358, opened 2023-02-09
feat: allow policy iam config to retrieve value using expression

rahmatrhd:
Changes:
- support expression evaluation in policy iam schema config
   example:
   ```json
   // response from identity service
   {"managers": [{"email": "user@example.com"}]}
   ```
   ```yaml
   # policy
   iam:
     ...
     schema:
       managers: $response.managers[0].email
   ```
   ```go
   // appeal result
   Appeal{
     ...
     Creator: map[string]interface{}{
       "managers": "user@example.com"
     },
   }
   ```

### DIFF
--- a/core/appeal/service.go
+++ b/core/appeal/service.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-playground/validator/v10"
 	"github.com/odpf/guardian/core/grant"
 	"github.com/odpf/guardian/domain"
+	"github.com/odpf/guardian/pkg/evaluator"
 	"github.com/odpf/guardian/plugins/notifiers"
 	"github.com/odpf/guardian/utils"
 	"github.com/odpf/salt/log"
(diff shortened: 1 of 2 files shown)


core/appeal/service.go
```diff
@@ -1017,7 +1019,18 @@ func (s *Service) addCreatorDetails(a *domain.Appeal, p *domain.Policy) error {
...
+							"response": userDetailsMap,
+						}
+						v, err := evaluator.Expression(targetKey).EvaluateWithVars(params)
+						if err != nil {
+							return fmt.Errorf("evaluating expression: %w", err)
+						}
+						creator[schemaKey] = v
+					} else {
+						creator[schemaKey] = userDetailsMap[targetKey]
+					}
```
bsushmith, line 1033: Can we simplify this block 😄 , the level of nesting in this block starting from `if userDetailsMap, ok := userDetails.(map[string]interface{}); ok` check is quite high. Maybe return early in starting flow?

bsushmith (approved):

PR #359, opened 2023-02-10
feat: improve performance

haveiss:
### Background
- The function `getActiveGrantsMap` retrieves all active grants, resulting in slow performance when a large number of records are returned.
- Some queries are slows due to not having an index

### Changes
- refactored to filter existing active grants in the database instead of retrieving all active data and filtering it on the application side.
- add indexes for slow queries

### DIFF
--- a/core/appeal/service.go
+++ b/core/appeal/service.go
@@ -222,7 +222,7 @@ func (s *Service) Create(ctx context.Context, appeals []*domain.Appeal, opts ...
 		}
 
 		activeGrant, err := s.findActiveGrant(ctx, appeal)
-		if err != nil {
+		if err != nil && err != ErrGrantNotFound {
 			return err
 		}
 
@@ -339,7 +339,7 @@ func (s *Service) findActiveGrant(ctx context.Context, a *domain.Appeal) (*domai
 	}
 
 	if len(grants) == 0 {
-		return nil, nil
+		return nil, ErrGrantNotFound
 	}
 
 	return &grants[len(grants)-1], nil
--- a/core/appeal/service.go
+++ b/core/appeal/service.go
@@ -332,6 +332,7 @@ func (s *Service) findActiveGrant(ctx context.Context, a *domain.Appeal) (*domai
 		AccountIDs:  []string{a.AccountID},
 		ResourceIDs: []string{a.ResourceID},
 		Roles:       []string{a.Role},
+		OrderBy:     []string{"updated_at:desc"},
 	})
 
 	if err != nil {
@@ -342,7 +343,7 @@ func (s *Service) findActiveGrant(ctx context.Context, a *domain.Appeal) (*domai
 		return nil, ErrGrantNotFound
 	}
 
-	return &grants[len(grants)-1], nil
+	return &grants[0], nil
 }
 
 func addOnBehalfApprovedNotification(appeal *domain.Appeal, notifications []domain.Notification) []domain.Notification {
--- a/core/appeal/service.go
+++ b/core/appeal/service.go
@@ -940,7 +940,7 @@ func (s *Service) GrantAccessToProvider(ctx context.Context, a *domain.Appeal, o
 	return nil
 }
 
-func (s *Service) checkExtensionEligibility(a *domain.Appeal, p *domain.Provider, policy *domain.Policy, grant *domain.Grant) error {
+func (s *Service) checkExtensionEligibility(a *domain.Appeal, p *domain.Provider, policy *domain.Policy, activeGrant *domain.Grant) error {
 	AllowActiveAccessExtensionIn := ""
 
 	// Default to use provider config if policy config is not set
@@ -964,7 +964,7 @@ func (s *Service) checkExtensionEligibility(a *domain.Appeal, p *domain.Provider
 		return fmt.Errorf("%w: %v: %v", ErrAppealInvalidExtensionDuration, AllowActiveAccessExtensionIn, err)
 	}
 
-	if !grant.IsEligibleForExtension(extensionDurationRule) {
+	if !activeGrant.IsEligibleForExtension(extensionDurationRule) {
 		return ErrGrantNotEligibleForExtension
 	}
 	return nil
--- a/core/appeal/service.go
+++ b/core/appeal/service.go
@@ -194,10 +194,6 @@ func (s *Service) Create(ctx context.Context, appeals []*domain.Appeal, opts ...
 	if err != nil {
 		return fmt.Errorf("listing pending appeals: %w", err)
 	}
-	activeGrants, err := s.getActiveGrantsMap(ctx)
-	if err != nil {
-		return fmt.Errorf("listing active grants: %w", err)
-	}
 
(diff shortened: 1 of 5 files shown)


core/appeal/service.go
```diff
@@ -225,8 +221,15 @@ func (s *Service) Create(ctx context.Context, appeals []*domain.Appeal, opts ...
...
 		}
 
-		if err := s.checkExtensionEligibility(appeal, provider, policy, activeGrants); err != nil {
-			return fmt.Errorf("checking grant extension eligibility: %w", err)
+		activeGrant, err := s.findActiveGrant(ctx, appeal)
+		if err != nil {
+			return err
+		}
+
+		if activeGrant != nil {
```
mabdh, line 229: is it the expected behavior if `activeGrant` is nil (no active grant found), the flow is still valid?

haveiss: yes, the flow should not changed other than the query, we have the existing test cases to cover this 
https://github.com/odpf/guardian/blob/main/core/appeal/service_test.go

core/appeal/service.go
```diff
@@ -323,6 +326,25 @@ func (s *Service) Create(ctx context.Context, appeals []*domain.Appeal, opts ...
...
+		AccountIDs:  []string{a.AccountID},
+		ResourceIDs: []string{a.ResourceID},
+		Roles:       []string{a.Role},
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("listing active grants: %w", err)
+	}
+
+	if len(grants) == 0 {
```
mabdh, line 341: If we expect this to return a single item, wouldn't it better to just return error not found here?

core/appeal/service.go
```diff
@@ -939,12 +939,7 @@ func (s *Service) GrantAccessToProvider(ctx context.Context, a *domain.Appeal, o
 	return nil
 }
 
-func (s *Service) checkExtensionEligibility(a *domain.Appeal, p *domain.Provider, policy *domain.Policy, activeGrants map[string]map[string]map[string]*domain.Grant) error {
-	grant, exists := activeGrants[a.AccountID][a.ResourceID][a.Role]
-	if !exists || grant == nil {
-		return nil
-	}
-
+func (s *Service) checkExtensionEligibility(a *domain.Appeal, p *domain.Provider, policy *domain.Policy, grant *domain.Grant) error {
```
rahmatrhd, line 942:
```suggestion
func (s *Service) checkExtensionEligibility(a *domain.Appeal, p *domain.Provider, policy *domain.Policy, activeGrant *domain.Grant) error {
```

core/appeal/service.go
```diff
@@ -323,6 +326,25 @@ func (s *Service) Create(ctx context.Context, appeals []*domain.Appeal, opts ...
...
+
+	if err != nil {
+		return nil, fmt.Errorf("listing active grants: %w", err)
+	}
+
+	if len(grants) == 0 {
+		return nil, nil
+	}
+
+	return &grants[len(grants)-1], nil
```
rahmatrhd, line 345: IMO adding `OrderBy=created_at` field in the filter makes `[len(grants)-1]` even more reasonable

core/appeal/service.go
```diff
@@ -323,6 +326,25 @@ func (s *Service) Create(ctx context.Context, appeals []*domain.Appeal, opts ...
 	return nil
 }
 
+func (s *Service) findActiveGrant(ctx context.Context, a *domain.Appeal) (*domain.Grant, error) {
+	grants, err := s.grantService.List(ctx, domain.ListGrantsFilter{
+		Statuses:    []string{string(domain.GrantStatusActive)},
+		AccountIDs:  []string{a.AccountID},
+		ResourceIDs: []string{a.ResourceID},
+		Roles:       []string{a.Role},
```
rahmatrhd, line 334: filtering active grants by `Role` should be avoided since we add `Permissions` into appeal (https://github.com/odpf/guardian/issues/205), can we also update this filter to use `Permissions` instead of `Role`? 

Note: this might resulting an increase in the query time since permissions column type is an array of string

internal/store/postgres/migrations/000014_add_indexes_to_improve_slow_queries.up.sql
```diff
@@ -0,0 +1,24 @@
+CREATE INDEX IF NOT EXISTS idx_appeals_created_by_status_deleted_at_updated_at
+ON appeals (created_by, status, deleted_at, updated_at DESC)
```
rahmatrhd, line 2: why are these fields indexed in a single index instead of creating individual index for each?

internal/store/postgres/migrations/000014_add_indexes_to_improve_slow_queries.up.sql
```diff
@@ -0,0 +1,24 @@
...
+WHERE deleted_at IS NULL;
+
+CREATE INDEX IF NOT EXISTS idx_grants_status_deleted_at_null
+ON grants (status, deleted_at)
+WHERE deleted_at IS NULL and status = 'active';
+
+CREATE INDEX IF NOT EXISTS idx_resources_id_deleted_at_null
+ON resources (id, deleted_at)
+WHERE deleted_at IS NULL;
+
```
rahmatrhd, line 24: can you add description comment on top of each index for better understanding?

rahmatrhd (changes requested):

core/appeal/service.go
bsushmith: I think, we can add another filter option for `permissions` if needed, but we should still have filtering at `role` level. Since account_id, resource_id, role are supposed to be a unique combination. 
I feel that adding permission here may bring out some false positives since some permissions can be common across different roles, and we would need to add checks for that again

rahmatrhd: the thing with role is that it can be updated in the provider config, so it is possible for an existing active grant and an appeal **with the same role id are having different set of permissions**. 

Using only permissions to filter active grant in this case would tells whether the user already have those permissions in the provider (regardless of the selected role id).

haveiss: current code are using this filter, i'm avoiding changing logic other than performance improvement in this PR
we can create separate PR for that, wdyt?
https://github.com/odpf/guardian/blob/2db7c75325dcaf28871dfd8aaa3956fa878272ab/core/appeal/service.go#L943-L946

internal/store/postgres/migrations/000014_add_indexes_to_improve_slow_queries.up.sql
haveiss: In general, it's a trade-off between index size and query performance. An index that covers multiple fields will be larger and slower to update, but can improve query performance. An index on a single field will be smaller and faster to update, but may not improve query performance as much.

IMHO better to use multiple fields for this case

core/appeal/service.go
rahmatrhd: yes this definitely can be detached to separate issue 👍

rahmatrhd (approved):
LGTM

PR #370, opened 2023-02-24
feat:  update slack notification by adding block functionality to template

utsav14nov:
feat:  update slack notification by adding block functionality to template


plugins/notifiers/slack/client.go
```diff
@@ -137,23 +142,23 @@ func (n *notifier) sendRequest(req *http.Request) (*userResponse, error) {
...
 	case domain.NotificationTypeApproverNotification:
-		text = templates.ApproverNotification
+		block = templates.ApproverNotification
 	case domain.NotificationTypeExpirationReminder:
-		text = templates.ExpirationReminder
+		block = templates.ExpirationReminder
 	case domain.NotificationTypeOnBehalfAppealApproved:
-		text = templates.OthersAppealApproved
+		block = templates.OthersAppealApproved
 	}
```
haveiss, line 159: hi, just have some thought, to improve readability rather than use switch how about to use map instead?

```go
	templateMap := map[string]string{
		domain.NotificationTypeAccessRevoked:          templates.AccessRevoked,
		...
	}

	text, ok := templateMap[message.Type]
	if !ok {
		return "", fmt.Errorf("template not found for message type %s", message.Type)
	}
```

plugins/notifiers/slack/client.go
```diff
@@ -67,11 +67,16 @@ func (n *notifier) Notify(items []domain.Notification) []error {
 	return errs
 }
 
-func (n *notifier) sendMessage(channel, text string) error {
+func (n *notifier) sendMessage(channel, block string) error {
```
bsushmith, line 70: would suggest to keep the variable name something like `messageBlock` or `message`. `block` as standalone word doesn't convey much in the func name.

core/appeal/service.go
```diff
@@ -797,6 +797,11 @@ func getApprovalNotifications(appeal *domain.Appeal) []domain.Notification {
 						"requestor":     appeal.CreatedBy,
 						"appeal_id":     appeal.ID,
 						"account_id":    appeal.AccountID,
+						"provider_type": appeal.Resource.ProviderType,
+						"resource_type": appeal.Resource.Type,
+						"created_at":    appeal.CreatedAt,
+						"approval_step": approval.Name,
+						"actor":         approver,
```
bsushmith, line 804: Can you add `appeals.details` also ?

utsav14nov: Done

plugins/notifiers/slack/client.go
utsav14nov: Done

plugins/notifiers/slack/client.go
utsav14nov: Done

plugins/notifiers/slack/client.go
```diff
@@ -137,23 +143,21 @@ func (n *notifier) sendRequest(req *http.Request) (*userResponse, error) {
...
+	messageTypeTemplateMap := map[string]string{
+		domain.NotificationTypeAccessRevoked:          templates.AccessRevoked,
+		domain.NotificationTypeAppealApproved:         templates.AppealApproved,
+		domain.NotificationTypeAppealRejected:         templates.AppealRejected,
+		domain.NotificationTypeApproverNotification:   templates.ApproverNotification,
+		domain.NotificationTypeExpirationReminder:     templates.ExpirationReminder,
+		domain.NotificationTypeOnBehalfAppealApproved: templates.OthersAppealApproved,
+	}
+
+	block, ok := messageTypeTemplateMap[message.Type]
```
bsushmith, line 155: can you change the variable name here too - `block` ?

plugins/notifiers/slack/client.go
```diff
@@ -67,11 +68,16 @@ func (n *notifier) Notify(items []domain.Notification) []error {
...
 }
 
-func (n *notifier) sendMessage(channel, text string) error {
+func (n *notifier) sendMessage(channel, messageBlock string) error {
 	url := slackHost + "/api/chat.postMessage"
-	data, err := json.Marshal(map[string]string{
+	var messageblockList []interface{}
+
+	if err := json.Unmarshal([]byte(messageBlock), &messageblockList); err != nil {
+		return err
```
bsushmith, line 76: Would suggest to add some context to the errors returned here, so that its easier to debug which place the code gave out

domain/notifier.go
```diff
@@ -1,12 +1,12 @@
...
+	AppealApproved       string `mapstructure:"appeal_approved" default:"[{\"type\":\"section\",\"text\":{\"type\":\"mrkdwn\",\"text\":\"Your appeal to {{.resource_name}} with role {{.role}} has been approved\"}}]"`
+	AppealRejected       string `mapstructure:"appeal_rejected" default:"[{\"type\":\"section\",\"text\":{\"type\":\"mrkdwn\",\"text\":\"Your appeal to {{.resource_name}} with role {{.role}} has been rejected\"}}]"`
+	AccessRevoked        string `mapstructure:"access_revoked" default:"[{\"type\":\"section\",\"text\":{\"type\":\"mrkdwn\",\"text\":\"Your access to {{.resource_name}}} with role {{.role}} has been revoked\"}}]"`
+	ApproverNotification string `mapstructure:"approver_notification" default:"[{\"type\":\"section\",\"text\":{\"type\":\"mrkdwn\",\"text\":\"You have an appeal created by {{.requestor}} requesting access to {{.resource_name}} with role {{.role}}.\\n Appeal ID: {{.appeal_id}}\"}}]"`
+	OthersAppealApproved string `mapstructure:"others_appeal_approved" default:"[{\"type\":\"section\",\"text\":{\"type\":\"mrkdwn\",\"text\":\"Your appeal to {{.resource_name}} with role {{.role}} created by {{.requestor}} has been approved\"}}]"`
```
bsushmith, line 9: I'm not getting good vibes(for lack of a better word) with how the default value is configured in here. 
@rahmatrhd @haveiss what's your opinion on this? Is this fine or do we have to think of using a constructer?

core/appeal/service.go
```diff
@@ -797,6 +797,13 @@ func getApprovalNotifications(appeal *domain.Appeal) []domain.Notification {
 						"requestor":     appeal.CreatedBy,
 						"appeal_id":     appeal.ID,
 						"account_id":    appeal.AccountID,
+						"account_type":  appeal.AccountType,
+						"provider_type": appeal.Resource.ProviderType,
+						"resource_type": appeal.Resource.Type,
+						"created_at":    appeal.CreatedAt,
+						"approval_step": approval.Name,
+						"actor":         approver,
+						"details":       appeal.Details,
```
rahmatrhd, line 806: how about adding `appeal` object, `approval` object, and `approver` string and allow user to use expression ta access them? while the existing fields are kept for backward compatibility

domain/notifier.go
rahmatrhd: for the default values, we can create .json files and programmatically set them as default values here.
but for actual customization in production I don't think there would be a better way to set the value since it'll be retrieved from env vars 🤔

plugins/notifiers/slack/client.go
```diff
@@ -137,23 +143,21 @@ func (n *notifier) sendRequest(req *http.Request) (*userResponse, error) {
...
+		domain.NotificationTypeOnBehalfAppealApproved: templates.OthersAppealApproved,
+	}
+
+	block, ok := messageTypeTemplateMap[message.Type]
+	if !ok {
+		return "", fmt.Errorf("template not found for message type %s", message.Type)
 	}
 
-	t, err := template.New("notification_messages").Parse(text)
+	t, err := template.New("notification_messages").Parse(block)
```
rahmatrhd, line 160: let's start using expression instead of template for better customizability?

domain/notifier.go
bsushmith: Agreed. Lets use .json files in code and set default. 

For deployment, it should be fine from using env vars, as the type is string. And as long as the user can maintain these strings while deploying, it should not be an issue to fetch from env vars.

utsav14nov:
@rahmatrhd As checked in terraform, we are be able to read full content from a json file as string and use it as variable in config which makes easily maintainable and structured code in production.
And as you suggested we can add the default config as separate json in guardian and read it programmatically through code.
What say? @bsushmith @haveiss

singhvikash11:
@rahmatrhd please have a look of repo https://github.com/odpf/guardian/pull/330

core/appeal/service.go
utsav14nov: This change is needed if we are going with the expression change.
But as checked , facing some error while using expression for templates, Will take it in a separate Pr.

plugins/notifiers/slack/client.go
utsav14nov: Done

plugins/notifiers/slack/client.go
utsav14nov: Done

domain/notifier.go
utsav14nov: For default templates, created templates folder with all the json inside it and used it programmatically in code.

plugins/notifiers/slack/client.go
utsav14nov: As checked, facing some errors.
Will take sometime to implement this change, therefore will take this in separate Pr after this Pr.

domain/notifier.go
```diff
@@ -1,12 +1,12 @@
...
-	AccessRevoked        string `mapstructure:"access_revoked" default:"[{\"type\":\"section\",\"text\":{\"type\":\"mrkdwn\",\"text\":\"Your access to {{.resource_name}}} with role {{.role}} has been revoked\"}}]"`
-	ApproverNotification string `mapstructure:"approver_notification" default:"[{\"type\":\"section\",\"text\":{\"type\":\"mrkdwn\",\"text\":\"You have an appeal created by {{.requestor}} requesting access to {{.resource_name}} with role {{.role}}.\\n Appeal ID: {{.appeal_id}}\"}}]"`
-	OthersAppealApproved string `mapstructure:"others_appeal_approved" default:"[{\"type\":\"section\",\"text\":{\"type\":\"mrkdwn\",\"text\":\"Your appeal to {{.resource_name}} with role {{.role}} created by {{.requestor}} has been approved\"}}]"`
+	ExpirationReminder   string `mapstructure:"expiration_reminder"`
+	AppealApproved       string `mapstructure:"appeal_approved" default:""`
+	AppealRejected       string `mapstructure:"appeal_rejected" default:""`
+	AccessRevoked        string `mapstructure:"access_revoked" default:""`
+	ApproverNotification string `mapstructure:"approver_notification" default:""`
+	OthersAppealApproved string `mapstructure:"others_appeal_approved" default:""`
```
bsushmith, line 9:
```suggestion
	ExpirationReminder   string `mapstructure:"expiration_reminder"`
	AppealApproved       string `mapstructure:"appeal_approved"`
	AppealRejected       string `mapstructure:"appeal_rejected"`
	AccessRevoked        string `mapstructure:"access_revoked"`
	ApproverNotification string `mapstructure:"approver_notification"`
	OthersAppealApproved string `mapstructure:"others_appeal_approved"`
```

plugins/notifiers/slack/client.go
```diff
@@ -148,6 +150,16 @@ func (n *notifier) sendRequest(req *http.Request) (*userResponse, error) {
 	return &result, nil
 }
 
+func getDefaultTemplate(messageType string) (string, error) {
+	pwd, _ := os.Getwd()
+	fullFilePath := fmt.Sprintf("%s/plugins/notifiers/slack/templates/%s.json", pwd, messageType)
+	content, err := ioutil.ReadFile(fullFilePath)
```
bsushmith, line 156: Why not have something like below? 🤔

bsushmith:
```suggestion
	bytes, err := os.ReadFile("notification.json")
```

domain/notifier.go
utsav14nov: Done

plugins/notifiers/slack/templates/AppealApproved.json
```diff
@@ -0,0 +1,9 @@
+[
+  {
+    "type":"section",
+    "text":{
+      "type":"mrkdwn",
+      "text":"Your appeal to {{.resource_name}} with role {{.role}} has been approved"
```
haveiss, line 6: let's add formatting to the value either make it bold or use code (backtick)
something like this

Your appeal for `{{.resource_name}}` with role `{{.role}}` has been approved.

plugins/notifiers/slack/client.go
```diff
@@ -136,24 +149,38 @@ func (n *notifier) sendRequest(req *http.Request) (*userResponse, error) {
 	return &result, nil
 }
 
+func getDefaultTemplate(messageType string) (string, error) {
+	content, err := os.ReadFile(fmt.Sprintf("plugins/notifiers/slack/templates/%s.json", messageType))
```
haveiss, line 153: should we [embed](https://pkg.go.dev/embed) the files or we add them when building the docker image?

```go
//go:embed plugins/notifiers/slack/templates/%s.json
var templateFiles embed.FS
```

plugins/notifiers/slack/client.go
```diff
@@ -43,7 +47,12 @@ type Config struct {
 }
 
 func New(config *Config) *notifier {
-	return &notifier{config.AccessToken, map[string]string{}, config.Messages}
+	return &notifier{
```
haveiss, line 50: shouldn't we initialized the default template by reading the json here?

plugins/notifiers/slack/client_test.go
```diff
@@ -0,0 +1,67 @@
...
+		Messages:     s.messages,
+		httpClient:   s.mockHttpClient,
+	}
+}
+
+func TestClient(t *testing.T) {
+	suite.Run(t, new(ClientTestSuite))
+}
+
+func (s *ClientTestSuite) TestNotify() {
```
haveiss, line 43: can we add positive case to cover reading the template file to work as expected?

core/appeal/service.go
rahmatrhd: @utsav14nov please create an issue for that

plugins/notifiers/slack/client.go
```diff
@@ -67,11 +76,16 @@ func (n *notifier) Notify(items []domain.Notification) []error {
...
 }
 
-func (n *notifier) sendMessage(channel, text string) error {
+func (n *notifier) sendMessage(channel, messageBlock string) error {
 	url := slackHost + "/api/chat.postMessage"
-	data, err := json.Marshal(map[string]string{
+	var messageblockList []interface{}
+
+	if err := json.Unmarshal([]byte(messageBlock), &messageblockList); err != nil {
+		return fmt.Errorf("error in parsing message block %s", err)
```
rahmatrhd, line 84: if the json unmarshal returns an error let's consider it as a normal string and send to slack using `text` field instead of `blocks`. This also keeps this change backward compatible

plugins/notifiers/slack/client_test.go
utsav14nov: Done

plugins/notifiers/slack/client.go
utsav14nov: Done

plugins/notifiers/slack/client.go
utsav14nov: Done

plugins/notifiers/slack/templates/AppealApproved.json
utsav14nov: Done

plugins/notifiers/slack/client.go
utsav14nov: Isn't it good to know the error than sending it as text.?
As we would be aware that there is something wrong in configuring notification.

We already has added the documentation for this change.

plugins/notifiers/slack/client_test.go
```diff
@@ -0,0 +1,77 @@
+package slack
```
haveiss, line 1: Let's use test package name like the others

```suggestion
package slack_test
```

haveiss (approved):

plugins/notifiers/slack/client_test.go
bsushmith: Will take these test package changes in a separate PR, since it would need some refactoring in the `slack/client.go` file to make `notifier` struct public.

bsushmith (approved):

PR #373, opened 2023-03-07
refactor: make slack.Notifier public

bsushmith:
* make `slack.Notifier` public

Changes for discussion - https://github.com/odpf/guardian/pull/370#discussion_r1125886870

### DIFF
--- a/plugins/notifiers/slack/client.go
+++ b/plugins/notifiers/slack/client.go
@@ -49,10 +49,10 @@ type Config struct {
 //go:embed templates/*
 var defaultTemplates embed.FS
 
-func NewNotifier(config *Config, slackIDCache map[string]string, httpClient utils.HTTPClient) *Notifier {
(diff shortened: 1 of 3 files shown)


plugins/notifiers/slack/client.go
```diff
@@ -51,25 +49,25 @@ type Config struct {
 //go:embed templates/*
 var defaultTemplates embed.FS
 
-func New(config *Config) *notifier {
-	return &notifier{
+func NewNotifier(config *Config, slackIDCache map[string]string, httpClient utils.HTTPClient) *Notifier {
```
rahmatrhd, line 52: can we exclude `slackIDCache` from `NewNotifier` parameter? this should be private for `Notifier` only

rahmatrhd (changes requested):

plugins/notifiers/slack/client.go
bsushmith: Done

rahmatrhd (approved):

PR #374, opened 2023-03-07
feat(grant): introduce update grant API

### LINKED ISSUE #362: Add API to assign/update owner of a grant
### Problem
With the recent [Import Grants](https://github.com/odpf/guardian/issues/206) feature, the default value for non-human access grants' `owner` field is empty. This is because Guardian can't determine the owner of the imported grants except for human accounts (individual email).

### Requirements
- Ability to assign a new owner to grants with empty owner
- Ability to replace the existing owner

### Proposed solution
1. Add API for updating grant's owner
   `PUT /grants/:id/owners` with request body: `{ "account_id": "user@example.com" }`
2. Add generic update grant API (preferred)
   `PATCH /grants/:id` with request body: `{ "owner": "user@example.com" }`
   - only allow field `owner`

#### Checklist
- [ ] Validate the owner value should not be empty
- [ ] Send notification to the new owner
- [ ] Send notification to the previous owner (if any)
- [ ] Record in audit log

### Next:
1. Introduce `co-owners: []string` field into Grant. An access/grant can be owned by multiple people especially if it's a service account or group access. `co-owners` will receive notifications like expiring notif and revocation notif the same way as the original owner.

rahmatrhd:
closes #362


core/grant/service.go
```diff
@@ -100,6 +101,66 @@ func (s *Service) GetByID(ctx context.Context, id string) (*domain.Grant, error)
...
+				"grant_id":       grantDetails.ID,
+				"previous_owner": previousOwner,
+				"new_owner":      updatedGrant.Owner,
+			},
+		}
+		notifications := []domain.Notification{{
+			User:    updatedGrant.Owner,
+			Message: message,
+		}}
+		if previousOwner != "" {
```
mabdh, line 148: should we check this if empty first? else we check if != updatedGrant owner

mabdh (approved):

PR #390, opened 2023-09-21
fix: bug in appeal creation

bsushmith:
- bug: If a final approval step of the policy gets skipped or allow_failed is true, then the appeal gets auto_approved, but the grant will not get created and no access is provided.
- fix bug by adding a check for last step
- add test-short cmd to makefile to be able to run quick service layer tests by skipping docker led store test cases.


AkarshSatija (approved):

ravisuhag:
@bsushmith test are failing for this.

ravisuhag:
@bsushmith fixed. It was a typo in makefile.

PR #395, opened 2023-11-15
refactor: frontier provider and add api documentation

Chief-Rishab:


plugins/providers/frontier/client.go
```diff
@@ -232,155 +238,170 @@ func (c *client) GetOrganizations() ([]*Organization, error) {
...
-	req, err := c.newRequest(http.MethodPost, endPoint, body, "")
+	req, err := c.newRequest(http.MethodPost, createPolicyEndpoint, body, "")
 	if err != nil {
 		return err
 	}
 
-	var users []*User
 	var response interface{}
 	if _, err := c.do(req, &response); err != nil {
+		c.logger.Error("Failed to grant access to the user", "Users", userId, req.URL)
```
kushsharma, line 254: Isn't the log format is `msg, key, val, key, val`? In this `req.URL` is set as `key`, it should have been
`c.logger.Error("Failed to grant access to the user", "Users", userId, "url", req.URL)`? Not sure, please confirm.

plugins/providers/frontier/client.go
```diff
@@ -232,155 +238,170 @@ func (c *client) GetOrganizations() ([]*Organization, error) {
...
-	var success successAccess
+	var policies []*Policy
 	var response interface{}
 	if _, err := c.do(req, &response); err != nil {
 		return err
 	}
 
 	if v, ok := response.(map[string]interface{}); ok && v != nil {
-		err = mapstructure.Decode(v, &success)
+		err = mapstructure.Decode(v[policiesConst], &policies)
```
kushsharma, line 318: what's the benefit of passing the `response` as `interface{}` first and then doing `Decode` again instead of directly decoding it in the actual type?

plugins/providers/frontier/client.go
```diff
@@ -232,155 +238,170 @@ func (c *client) GetOrganizations() ([]*Organization, error) {
...
-	c.logger.Info("Organization access to the user,", "total users", len(users), req.URL)
+	c.logger.Info("Organization access to the user,", "Users", userId, req.URL)
 	return nil
 }
 
-func (c *client) RevokeTeamAccess(resource *Team, userId string, role string) error {
-	endPoint := path.Join(groupsEndpoint, "/", resource.ID, "/", role, "/", userId)
-	req, err := c.newRequest(http.MethodDelete, endPoint, "", "")
+func (c *client) RevokeGroupAccess(resource *Group, userId string, role string) error {
+	endpoint := createPolicyEndpoint + "?groupId=" + resource.ID + "&userId=" + userId + "&roleId=" + role
```
kushsharma, line 305: we can refactor this to use `url.URL` or `fmt.Sprintf`

kushsharma (commented):
Other than minor comments, LGTM

kushsharma (approved):

